PR #3, opened 2020-06-22
General fixes

fabriziovigevani:
- Moved proposal actions into its own hook
- Separated the logic for fetching the organization and the app data (This prevents refetching when account is connected)
- Adds stakeHistory data fetching
- Created subscription hooks so we can now subscribe to data instead of querying it one time 

Subscriptions are not yet working as expected, is currently performing a single query (will review tomorrow with @0xGabi)

This uses a new version of the connector so make sure to install the latest version.


src/app-logic.js
```diff
@@ -58,30 +33,31 @@ export default function useAppLogic() {
...
+          addressesEqual(stake.entity, account)
         )
 
         if (myStake) {
           myStakes.push({
-            proposal: proposal.id,
+            proposalId: proposal.id,
             proposalName: proposal.name,
-            stakedAmount: myStake.amount,
+            amount: myStake.amount,
```
fabriziovigevani, line 44: Note that decided to update this attributes to be consistent across all the app's data

src/hooks/useSubscriptions.js
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    if (!convictionVoting) {
+      return
+    }
+
+    proposalsSubscription.current = convictionVoting.onProposals(
+      onProposalsHandler
+    )
+
+    return () => proposalsSubscription.current[0]()
```
fabriziovigevani, line 26: The  `onPropsals` function is supposed to return the object with the `unsubscribe` function (see https://github.com/1Hive/conviction-voting-app/blob/bdd20ffa7110475d15eb36576fa727b3ba5c6e0a/connector/src/entities/ConvictionVoting.ts#L24) 
but it's actually returning an array where the first element is the unsubscribe function

.gitignore
```diff
@@ -2,6 +2,7 @@
 # dependencies
 /node_modules
 package-lock.json
+yarn.lock
```
rperez89, line 5: we might not want to upload the lock fie

src/app-logic.js
```diff
@@ -58,30 +33,31 @@ export default function useAppLogic() {
...
       { myStakes: [], totalActiveTokens: new BigNumber('0') }
     )
-  }, [proposals, connectedAccount, stakeToken])
+  }, [account, proposals, stakeToken])
+
+  const actions = useProposalActions(proposalPanel.requestClose)
 
   return {
+    actions,
+    isLoading: isLoading, // TODO: Add loading flag for block back again
```
rperez89, line 57: i think that now we are able to add that flag again right?

src/hooks/useAppHooks.js
```diff
@@ -1,13 +1,36 @@
 import { useEffect, useRef, useState } from 'react'
```
rperez89, line 1: Should we maybe rename this file to `useOrgHooks`? because we have a mix of differents apps and org hooks here, not sure but what do you think?

src/hooks/useAppHooks.js
```diff
@@ -23,6 +46,105 @@ query miniMeToken($id: String!, $address: String!) {
...
+}
+
+export function useOrganzation() {
+  const [organzation, setOrganization] = useState(null)
+  const { ethereum } = useWallet()
+
+  useEffect(() => {
+    let cancelled = false
+    const fetchOrg = async () => {
+      // Fetch the apps belonging to this organization. const apps = await organization.apps()
```
rperez89, line 67: I think this comment is wrong

src/hooks/useAppHooks.js
```diff
@@ -46,7 +168,8 @@ export function useVaultBalance(installedApps, token, timeout = 1000) {
           const vaultBalance = await vaultContract.balance(token.id)
 
           if (!cancelled) {
-            setVaultBalance(vaultBalance)
+            // Contract value is bn.js so we need to convert it to bignumber.js
```
rperez89, line 171: is always like this? doesn't the contract calls return a string?

src/hooks/useBlock.js
```diff
@@ -4,14 +4,16 @@ import useInterval from './useInterval'
 import { useWallet } from '../providers/Wallet'
 
 export function useLatestBlock(updateEvery = 1000) {
-  const { getBlockNumber } = useWallet()
+  const { ethers } = useWallet()
```
rperez89, line 7: i think we just need the `getBlock`

src/providers/AppState.js
```diff
@@ -1,111 +1,39 @@
...
-      cancelled = true
-    }
-  }, [])
+  const organization = useOrganzation()
+  const appData = useAppData(organization)
 
   const vaultBalance = useVaultBalance(
-    appState.installedApps,
-    appState.requestToken
+    appData.installedApps,
```
rperez89, line 20: should we do `{installedApps, requestToken, stakeToken, convictionVoting} = appData` above?

.gitignore
fabriziovigevani: Do you mean we **might want** to upload it? 

Decided to remove it for now because it gave me problems when i wanted to install updated dependencies and lost some time finding out was due to the lock file.

Maybe we could consider re adding it later once the code is more stable?

rperez89: ahh no my bad i thought that we were adding it, we are in gitignore haha , yes i think that we should not upload it

src/hooks/useBlock.js
fabriziovigevani: The `getBlock` function is part of  `ethers`.  `useWallet` provides the `getBlockNumber` function but couldn't make it work that's why i changed it

src/hooks/useAppHooks.js
fabriziovigevani: Thought it did return a string but it just returns a `BN`

src/app-logic.js
fabriziovigevani: Not yet, the blocks start to come when the account is enabled since it using the `ethers` from  `useWallet`. We need to resolve this by using our own created provider. Will work on it on a separate PR

src/hooks/useAppHooks.js
fabriziovigevani: Yeah that makes sense. And we will most likely be adding more data from different apps.

src/app-logic.js
rperez89: right!

PR #7, opened 2020-06-24
Proposal Detail : Features migrated from conviction-voting

rperez89:
This PR include the latest changes to the proposal screen on the conviction voting repo

### DIFF
--- a/src/hooks/useProposalActions.js
+++ b/src/hooks/useProposalActions.js
@@ -29,6 +29,7 @@ export default function useProposalActions(onDone) {
 
   const stakeToProposal = useCallback(
     (proposalId, amount) => {
+      console.log('STAKE ', typeof amount)
       sendIntent(
         organization,
         convictionVoting.appAddress,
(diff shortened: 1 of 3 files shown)


src/hooks/useProposalActions.js
```diff
@@ -29,6 +29,7 @@ export default function useProposalActions(onDone) {
 
   const stakeToProposal = useCallback(
     (proposalId, amount) => {
+      console.log('STAKE ', typeof amount)
```
fabriziovigevani, line 32:
```suggestion

```

src/hooks/useProposalActions.js
```diff
@@ -87,6 +88,10 @@ async function sendIntent(
 
     const txPath = await intent.paths(from)
 
+    console.log('txPath ', txPath)
+
+    console.log('PARAMS! ', params)
+
```
fabriziovigevani, line 94:
```suggestion

```

src/screens/ProposalDetail.js
```diff
@@ -40,7 +53,7 @@ function ProposalDetail({
 
   const { account: connectedAccount } = useWallet()
   const chartLines = useConvictionHistory(proposal)
-
+  const { stakeToken, accountBalance } = useAppState()
```
fabriziovigevani, line 56: Ah i think in this case the `stakeToken` is already passed from `App.js`

rperez89: i think that the one passed is re requestToken

PR #22, opened 2020-06-29
Bug fixes

fabriziovigevani:
- Fixed wrong attribute in `stakesHistory` entity.
- Created mapping for block times per network
- Increased 70k gas_limit for `execute` action

### DIFF
--- a/src/hooks/useActions.js
+++ b/src/hooks/useActions.js
@@ -4,7 +4,7 @@ import { useWallet } from '../providers/Wallet'
 import { toHex } from 'web3-utils'
 import { getAppAddressByName } from '../lib/data-utils'
 
-const GAS_LIMIT = 350000
+const GAS_LIMIT = 420000
 
 export default function useActions(onDone) {
   const { account, ethers } = useWallet()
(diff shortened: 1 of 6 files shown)


src/hooks/useActions.js
```diff
@@ -4,7 +4,7 @@ import { useWallet } from '../providers/Wallet'
 import { toHex } from 'web3-utils'
 import { getAppAddressByName } from '../lib/data-utils'
 
-const GAS_LIMIT = 350000
+const GAS_LIMIT = 420000
```
rperez89, line 7: given that with this chain we are having some issues regarding the gas limit, wondering if we should't use an ENV var to avoid deploying again on each gas issue that we might found

src/networks.js
```diff
@@ -17,8 +17,8 @@ const networks = {
   xdai: {
     chainId: 100,
     ensRegistry: '0xaafca6b0c89521752e559650206d7c925fd0e530',
-    name: 'xdai',
-    type: 'private',
+    name: 'xDai',
```
rperez89, line 20: i thin that greef suggested to rename it to xDAI

src/hooks/useActions.js
fabriziovigevani: I don't think an ENV var would help us skip the re deployment? Would make sense if we have different deployments for different chains tough.

src/networks.js
fabriziovigevani: Ah, it said that in the PR title but it's actually `xDai` https://github.com/1Hive/honey-pot/pull/19

PR #28, opened 2020-07-03
Uniswap price info

fabriziovigevani:
Fetches Honey price info from deployed Uniswap instance on xDai.
Screen Shot 2020-07-06 at 1 59 32 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,10 +5,11 @@
   "dependencies": {
     "@1hive/1hive-ui": "^1.0.1",
     "@1hive/connect-thegraph-conviction-voting": "^0.1.7",
+    "@1hive/uniswap-sdk": "^1.0.0",
     "@aragon/connect": "^0.3.0",
     "bignumber.js": "^9.0.0",
     "clipboard-polyfill": "^2.8.6",
-    "ethers": "^5.0.2",
+    "ethers": "4.0.28",
     "node-fetch": "^2.6.0",
     "react": "^16.13.1",
     "react-app-polyfill": "^1.0.6",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -5,10 +5,11 @@
   "dependencies": {
     "@1hive/1hive-ui": "^1.0.1",
     "@1hive/connect-thegraph-conviction-voting": "^0.1.7",
+    "@1hive/uniswap-sdk": "^1.0.0",
     "@aragon/connect": "^0.3.0",
     "bignumber.js": "^9.0.0",
     "clipboard-polyfill": "^2.8.6",
-    "ethers": "^5.0.2",
+    "ethers": "4.0.28",
```
fabriziovigevani, line 12: Bote that we needed to downgrade ethers version so it could work with uniswap v1 sdk

lkngtn:
I think we could replace "open proposals" with "honey price" (and maybe put honey price as the first entry). Maybe we can also put a small link bellow the price "trade" that links to the uniswap.1hive.org (we may change that when we have a better design but it seems useful for now).

src/hooks/useTokenPrice.js
```diff
@@ -0,0 +1,74 @@
...
+import { formatTokenAmount } from '../lib/token-utils'
+
+const UNISWAP_PRECISION = 18
+const UNISWAP_MARKET_RETRY_EVERY = 1000
+
+async function getTokenMarketDetails(tokenAddress, provider) {
+  const tokenReserves = await getTokenReserves(tokenAddress, provider)
+
+  // For XDAI we must pass undefined
+  return getMarketDetails(undefined, tokenReserves)
```
rperez89, line 14: Is not easy way to do the changes on the uniswap lib and pass another thing rather than undefined?

fabriziovigevani: Note that that's just how the sdk works. https://uniswap.org/docs/v1/SDK/computation/

For `ETH` you have to pass `undefined` as the token reserves.

PR #30, opened 2020-07-08
Medium border radius 

rperez89:
By @fioreb request we are using a border radius of 12 in some components.
Also fixed a radius problem on the new proposal panel

Screen Shot 2020-07-08 at 5 13 40 PM

Screen Shot 2020-07-08 at 5 13 19 PM

### DIFF
--- a/src/components/AddProposalPanel.js
+++ b/src/components/AddProposalPanel.js
@@ -5,6 +5,7 @@ import {
   GU,
   Info,
   isAddress,
+  MEDIUM_RADIUS,
   TextInput,
   useTheme,
 } from '@1hive/1hive-ui'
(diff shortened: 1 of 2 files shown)


src/components/AddProposalPanel.js
```diff
@@ -195,6 +195,7 @@ const AddProposalPanel = React.memo(({ onSubmit }) => {
                   css={`
                     background: ${theme.background};
                     border-left: 1px solid ${theme.border};
+                    border-radius: 0px 12px 12px 0px;
```
fabriziovigevani, line 198: Should we import `MEDIUM_RADIUS` and use it here?

PR #36, opened 2020-08-04
Update @aragon/connect to 0.5.2

bpierre:
[This release](https://github.com/aragon/connect/releases/tag/v0.5.2) fixes the issue with webpack mixing the two Ethers.js versions.

I also added a yarn.lock file to ensure the versions are locked, but let me know if I should remove it.


src/hooks/useOrgHooks.js
```diff
@@ -21,8 +21,6 @@ import {
 
 // Organzation
 const ORG_ADDRESS = '0xe9869a0bbc8fb8c61b7d81c33fa2ba84871b3b0e'
-const ORG_SUBRAPH_URL =
-  'https://api.thegraph.com/subgraphs/name/1hive/aragon-xdai'
```
bpierre, line 25: This is not needed since [Connect knows about xDai](https://github.com/aragon/connect/blob/master/packages/connect-tokens/src/thegraph/connector.ts#L16-L18), and it will also allow Connect to pick the right one when started with another network ([e.g. `4`](https://github.com/1Hive/honey-pot/blob/fd4fabb1928dac344a0c0ccdcf4763639fd23fce/package.json#L27)).

src/hooks/useOrgHooks.js
```diff
@@ -63,19 +61,10 @@ export function useOrganzation() {
...
-            orgSubgraphUrl: ORG_SUBRAPH_URL,
-          },
-        ],
-        {
-          readProvider: ethereum || ethers,
-          chainId: getDefaultChain(),
-        }
-      )
+      const organization = await connect(ORG_ADDRESS, 'thegraph', {
+        ethereum,
```
bpierre, line 65: The name for the provider is now `ethereum`. Also `ethers` should not be passed here, as it is not an Ethereum provider object.

src/hooks/useOrgHooks.js
```diff
@@ -63,19 +61,10 @@ export function useOrganzation() {
...
+        network: getDefaultChain(),
```
bpierre, line 66: `chainId` is now `network`. It can also now accept [a chain ID, a network name, or a network object](https://github.com/aragon/connect/blob/bf0eec72b9320d9166e1cd5717dc1f9f4a0b8899/packages/connect-types/src/index.ts#L9-L12).

src/hooks/useOrgHooks.js
rperez89: so you guys deployed an aragon xdai subgraph? because here we are using the one under `/1hive`

bpierre: Oh you’re right sorry, I checked too quickly and thought it was the same URL. I don’t know if there is a difference between the two. @0xGabi do you think we could use the Aragon one here?

0xGabi: Actually I configured it to use the one deployed by 1Hive 👍

PR #38, opened 2020-08-07
Account: Add 3box profile names

Evalir:
Integrates 3Box profile names in a very minimal way, defaulting to the shortened address when an user doesn't have a profile. Later we could look into adding comments for every proposal, albeit this requires quite more code. :)

Screenshot from my fork of the repo:
Screen Shot 2020-08-06 at 10 27 36 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "3box": "^1.21.0",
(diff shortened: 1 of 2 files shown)


lkngtn:
Will this also pull the profile image, if you have it set on 3box?

Evalir:
Nope, but we could replace the Identicon with the image in that case.

Evalir:
Added it and merged it!

PR #41, opened 2020-08-10
Cancel proposal

rperez89:
Needs the new Conviction voting smart contract to be deployed to xdai - rinkeby

Built on top https://github.com/1Hive/honey-pot/pull/40

Screen Shot 2020-08-10 at 4 36 57 PM

### DIFF
--- a/src/screens/MainScreen.js
+++ b/src/screens/MainScreen.js
@@ -11,9 +11,9 @@ const MainScreen = React.memo(
   ({
     isLoading,
     myStakes,
+    onCancelProposal,
     onExecuteIssuance,
     onExecuteProposal,
-    onCancelProposal,
     onRequestNewProposal,
     onStakeToProposal,
     onWithdrawFromProposal,
--- a/src/screens/MainScreen.js
+++ b/src/screens/MainScreen.js
@@ -13,6 +13,7 @@ const MainScreen = React.memo(
     myStakes,
     onExecuteIssuance,
     onExecuteProposal,
+    onCancelProposal,
     onRequestNewProposal,
     onStakeToProposal,
     onWithdrawFromProposal,
@@ -59,6 +60,7 @@ const MainScreen = React.memo(
           <ProposalDetail
             onBack={handleBack}
             onExecuteProposal={onExecuteProposal}
+            onCancelProposal={onCancelProposal}
             onStakeToProposal={onStakeToProposal}
             onWithdrawFromProposal={onWithdrawFromProposal}
             proposal={selectedProposal}
--- a/src/screens/ProposalDetail.js
+++ b/src/screens/ProposalDetail.js
@@ -31,13 +31,13 @@ import {
   soliditySha3,
 } from '../lib/web3-utils'
 
-const cancelRoleHash = soliditySha3('CANCEL_PROPOSAL_ROLE')
+const CANCEL_ROLE_HASH = soliditySha3('CANCEL_PROPOSAL_ROLE')
 
 function ProposalDetail({
   proposal,
   onBack,
-  onExecuteProposal,
   onCancelProposal,
+  onExecuteProposal,
   onStakeToProposal,
   onWithdrawFromProposal,
   requestToken,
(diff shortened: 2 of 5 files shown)


src/screens/MainScreen.js
```diff
@@ -59,6 +60,7 @@ const MainScreen = React.memo(
           <ProposalDetail
             onBack={handleBack}
             onExecuteProposal={onExecuteProposal}
+            onCancelProposal={onCancelProposal}
```
fabriziovigevani, line 63: Could we alphabetize it?

src/screens/ProposalDetail.js
```diff
@@ -25,12 +26,18 @@ import usePanelState from '../hooks/usePanelState'
...
+  soliditySha3,
+} from '../lib/web3-utils'
+
+const cancelRoleHash = soliditySha3('CANCEL_PROPOSAL_ROLE')
 
 function ProposalDetail({
   proposal,
   onBack,
   onExecuteProposal,
+  onCancelProposal,
```
fabriziovigevani, line 40: Same here

src/screens/ProposalDetail.js
```diff
@@ -25,12 +26,18 @@ import usePanelState from '../hooks/usePanelState'
 import { useWallet } from '../providers/Wallet'
 
 import { getTokenIconBySymbol, formatTokenAmount } from '../lib/token-utils'
-import { addressesEqualNoSum as addressesEqual } from '../lib/web3-utils'
+import {
+  addressesEqualNoSum as addressesEqual,
+  soliditySha3,
+} from '../lib/web3-utils'
+
+const cancelRoleHash = soliditySha3('CANCEL_PROPOSAL_ROLE')
```
fabriziovigevani, line 34: Let's rename it to `CANCEL_PROPOSAL_ROLE` as it is in the contract?

src/screens/ProposalDetail.js
```diff
@@ -39,9 +46,21 @@ function ProposalDetail({
...
+    if (!connectedAccount) {
+      return false
+    }
+    return permissions.find(
+      ({ roleHash, granteeAddress }) =>
+        roleHash === cancelRoleHash &&
+        addressesEqual(granteeAddress, connectedAccount)
+    )
+  }, [connectedAccount, permissions])
+
```
fabriziovigevani, line 63: Ah, let's rename the variable to `canCancelProposal` or something around that because the submitter of the proposal can cancel it too without the need of being granted the cancel role, so we should add that logic as well.

rperez89: ah nice

src/screens/ProposalDetail.js
rperez89: here i would prefer to make explicit with the name that is the hash and not the string so just reading the name you know what is it :)

fabriziovigevani: Ok that's fair. Maybe have it in upper case though as we do with all constants?

src/screens/ProposalDetail.js
```diff
@@ -71,6 +60,18 @@ function ProposalDetail({
...
+    if (!connectedAccount) {
+      return false
+    }
+    return permissions.find(
+      ({ roleHash, granteeAddress }) =>
+        (roleHash === CANCEL_ROLE_HASH &&
+          addressesEqual(granteeAddress, connectedAccount)) ||
+        addressesEqual(creator, connectedAccount)
+    )
+  }, [connectedAccount, creator, permissions])
```
fabriziovigevani, line 73: Would  refactor to something like:

```suggestion
  const canCancelProposal = useMemo(() => {
    if (!connectedAccount) {
      return false
    }
    
    if (addressesEqual(creator, connectedAccount)) {
      return true
    }
    
    return permissions.find(
      ({ roleHash, granteeAddress }) =>
        (roleHash === CANCEL_ROLE_HASH &&
          addressesEqual(granteeAddress, connectedAccount))
    )
  }, [connectedAccount, creator, permissions])
```

PR #47, opened 2020-08-12
Migration to parcel

### LINKED ISSUE #49: Use parcel bundler
Right now we are using the defualt react scritps with a huge webpack config file that is not the best.

rperez89:
This PR includes

- Migration to parcel bundler

- Build and start scripts

- Remove react scripts and dependencies

- Remove public folder

- Remove uniswap integration since is failing because of the ethers version issue

- Update ethers to v5

- Remove usd prices from wallet and metrics

closes #49


src/providers/Wallet.js
```diff
@@ -18,6 +18,7 @@ function WalletAugmented({ children }) {
 
   const ethers = useMemo(() => {
     if (!ethereum) {
+      console.log('get  network ', getNetwork())
```
fabriziovigevani, line 21:
```suggestion
```

PR #48, opened 2020-08-12
Update connector

fabriziovigevani:
- Updates @aragon/connect usage
- Updates conviction connector usage
- Fetches stake token balance and total supply via contract call as opposed to subgraph query.


src/hooks/useSubscriptions.js
```diff
@@ -47,6 +48,7 @@ export function useStakesHistorySubscription(convictionVoting) {
     }
 
     stakesSubscription.current = convictionVoting.onStakesHistory(
+      {},
```
fabriziovigevani, line 51: This was causing the previous `callback not a function` error

src/hooks/useOrgHooks.js
```diff
@@ -1,50 +1,27 @@
...
-    id
-    totalSupply
-    holders(where: {address: $address}){
-      address
-      balance
-    }
-  }
-}
-`
+const APP_NAME = 'conviction-beta' // TODO: save elsewhere
```
Evalir, line 24: I was thinking we should co-locate all these configuration related constants either in an `environment.js` file so we can change it with environment variables, or even lift them up into a context so we can change some things at runtime.

src/hooks/useOrgHooks.js
```diff
@@ -201,44 +167,55 @@ export function useVaultBalance(installedApps, token, timeout = 1000) {
...
-      const { miniMeToken } = results.data
+        const contractTotalSupply = await tokenContract.totalSupply()
 
-      if (!cancelled) {
-        setBalances({
-          balance: new BigNumber(miniMeToken.holders[0]?.balance || 0),
-          totalSupply: new BigNumber(miniMeToken.totalSupply),
-        })
+        if (!cancelled) {
+          // Contract value is bn.js so we need to convert it to bignumber.js
```
Evalir, line 196: Wondering why we're using bignumber.js here? Is it for the decimal utils?

rperez89: yeah, for example the total supply is used to do the maths to calculate the convction and threshold and since they are really big numbers  with lot of decimals we need it https://github.com/1Hive/honey-pot/blob/master/src/hooks/useProposals.js#L50-L62

Evalir: Aaah yeah, I meant why we're using `bignumber.js` rather than `bn.js` 😛

rperez89: ahh because it alow us to use .xxxxxxxxxxxx really big with bn.js was not accurate at all

Evalir: Ah gotcha! Makes sense. I was mainly saying it because the native BigInts that come on Chrome and Firefox (only Safari's behind :/ ) [only support integers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt), so I generally prefer `bn.js` or `JSBI` which is actually the closest to the native BigInt standard. That said though, having decimal support is super nice!

src/hooks/useOrgHooks.js
fabriziovigevani: Yep let's do it now!

fabriziovigevani: Went forward with adding env variables for the org address and app name.

Org address defaults to the ones specified at `networks.js` file if none sepcified
App name defaults to `conviction-voting` if none specified.

Evalir: Makes sense! :)

Evalir:
Just a thing: I noticed that with the `0.2.0` version of the connector, the app address is found with the key `address` instead of `appAddress` in the conviction voting object. This means the intents in `useActions` will fail and will need to be updated accordingly.

rperez89:
> Just a thing: I noticed that with the `0.2.0` version of the connector, the app address is found with the key `address` instead of `appAddress` in the conviction voting object. This means the intents in `useActions` will fail and will need to be updated accordingly.

ahh good catch!!

fabriziovigevani:
> This means the intents in useActions will fail and will need to be updated accordingly.

Updated. Thanks @Evalir !

PR #52, opened 2020-08-13
Remove proposal: Info message and filter

rperez89:
Screen Shot 2020-08-13 at 6 55 27 PM

Screen Shot 2020-08-13 at 6 55 44 PM


src/constants.js
```diff
@@ -1 +1,4 @@
+export const PROPOSAL_STATUS_ACTIVE = 'Active'
+export const PROPOSAL_STATUS_CANCELLED = 'Cancelled'
+export const PROPOSAL_STATUS_EXECUTED = 'Executed'
```
fabriziovigevani, line 3: Wondering if we should declare this constants inside the proposal-types file with the _STRING suffix?

src/screens/ProposalDetail.js
```diff
@@ -52,14 +57,16 @@ function ProposalDetail({
 
   const { account: connectedAccount } = useWallet()
 
+  console.log('PROPOSAL!!! ', proposal)
```
fabriziovigevani, line 60:
```suggestion
```

PR #54, opened 2020-08-16
Proposal Detail: Supporters distribution

rperez89:
Needed to do some adjustments to the designs and use smallers sizes given the small space that we have

Screen Shot 2020-08-17 at 6 46 25 PM


src/screens/ProposalDetail.js
```diff
@@ -84,6 +88,10 @@ function ProposalDetail({
 
   const signalingProposal = addressesEqual(beneficiary, ZERO_ADDR)
 
+  const filteredStakes = stakes.filter(({ amount }) => {
```
fabriziovigevani, line 91: Should we wrap it with a `useMemo` ?

src/components/SupportersDistribution.js
```diff
@@ -0,0 +1,162 @@
...
+        connectedAccount={isCurrentUser}
+        compact
+        css={`
+          width: ${compactMode ? 'auto' : '110px'};
+        `}
+      />
+      <div
+        css={`
+          display: flex;
+          float: right;
```
fabriziovigevani, line 134: Don't think the `float` prop is needed here. We are also missing the center alignment.

We could also move the `textStyle` directly to this parent node.

```suggestion
    align-items: center;
    ${textStyle('body4')};
```

src/components/SupportersDistribution.js
```diff
@@ -0,0 +1,162 @@
...
+        `}
+      >
+        <span
+          css={`
+            margin-right: ${0.5 * GU}px;
+            color: ${theme.contentSecondary};
+            ${textStyle('body4')};
```
fabriziovigevani, line 141: From previous  comment

```suggestion
```

src/components/SupportersDistribution.js
```diff
@@ -0,0 +1,162 @@
...
+          `}
+        >
+          {percentage}%
+        </span>
+
+        <span
+          css={`
+            ${textStyle('body4')};
```
fabriziovigevani, line 149: From previous comment

```suggestion
```

src/components/SupportersDistribution.js
```diff
@@ -0,0 +1,162 @@
...
+      `}
+    >
+      <IdentityBadge
+        entity={entity}
+        connectedAccount={isCurrentUser}
+        compact
+        css={`
+          width: ${compactMode ? 'auto' : '110px'};
+        `}
+      />
```
fabriziovigevani, line 130: Maybe we could also have a `body4` text style for the label so it's the same as the amount and % style?

```suggestion
      <IdentityBadge
        entity={entity}
        connectedAccount={isCurrentUser}
        compact
        labelStyle={`${textStyle('body4')}`};
        css={`
          width: ${compactMode ? 'auto' : '110px'};
        `}
      />
```

src/components/SupportersDistribution.js
```diff
@@ -0,0 +1,162 @@
...
+              </div>
+            )
+          }}
+        />
+        <div
+          css={`
+            margin-top: ${3 * GU}px;
+            display: flex;
+            justify-content: space-between;
+            ${textStyle('body4')}
```
fabriziovigevani, line 93: May make sense to increase the size of the `Total`

```suggestion
            ${textStyle('body3')}
```

src/components/SupportersDistribution.js
rperez89: why do you think that there is a need to align only the percentage and the honey value? it will be different from the label alignment

fabriziovigevani: Screen Shot 2020-08-17 at 7 42 41 PM

The label `div` wrapping the `%` and `amount` is stretched due to the `label` being bigger.
Aligning the items to the center would also align them indirectly with the label.

PR #58, opened 2020-08-23
Signaling , effective supply  and other fixes

rperez89:
This PR includes 

- some visual changes for when is a signaling proposal

- Empty stakers component 

- Upgrade to new version of "@1hive/connect-conviction-voting"

- Upgrade to new version of "@aragon/connect"

- Fix the polling issue 

- Get the new minThresholdStakePercentage, and totalStaked from the subgraph

- Calculate threshold based on the totalStaked, minThresholdStakePercentage and totaSupply

Screen Shot 2020-08-24 at 2 57 22 PM

Screen Shot 2020-08-24 at 2 54 19 PM


src/hooks/useProposals.js
```diff
@@ -97,6 +82,31 @@ export function useProposals() {
         TIME_UNIT
       )
 
+      // Funding proposal needed values
+      if (!proposal.requestedAmount.eq(0)) {
```
fabriziovigevani, line 86:
```suggestion
      if (proposal.requestedAmount.gt(0)) {
```

src/hooks/useSubscriptions.js
```diff
@@ -1,9 +1,33 @@
 import { useCallback, useEffect, useRef, useState } from 'react'
 import {
+  transformConfigData,
   transformProposalData,
   transformStakeHistoryData,
 } from '../lib/data-utils'
 
+export function useConfigSubscription(convictionVoting) {
+  const [config, setConfig] = useState([])
```
fabriziovigevani, line 9: should be `null` right?

```suggestion
  const [config, setConfig] = useState(null)
```

src/providers/AppState.js
```diff
@@ -19,15 +20,34 @@ function AppStateProvider({ children }) {
     installedApps,
     requestToken,
     stakeToken,
+    totalStaked,
+    minThresholdStakePercentage,
+    pctBase,
```
fabriziovigevani, line 25: Could we alphabetize them?

src/providers/AppState.js
```diff
@@ -19,15 +20,34 @@ function AppStateProvider({ children }) {
...
 
+  const effectiveSupply = useMemo(() => {
+    if (
+      !(totalSupply && totalStaked && minThresholdStakePercentage && pctBase)
+    ) {
+      return
+    }
+    const percentageOfTotalSupply = totalSupply
+      .multipliedBy(minThresholdStakePercentage)
+      .div(bigNum(1))
```
fabriziovigevani, line 41: Let's maybe create a new const in `./constants.js` namedsomething like `STAKE_PCT_BASE`.

src/hooks/useOrgHooks.js
```diff
@@ -104,13 +102,14 @@ export function useAppData(organization) {
...
 
+  const config = useConfigSubscription(appData.convictionVoting)
   const proposals = useProposalsSubscription(appData.convictionVoting)
 
   // Stakes done across all proposals on this app
   // Includes old and current stakes
   const stakesHistory = useStakesHistorySubscription(appData.convictionVoting)
 
-  return { ...appData, proposals, stakesHistory }
+  return { ...appData, ...config, proposals, stakesHistory }
```
fabriziovigevani, line 112: Wondering if makes sense to have all config values inside the `config` object instead of spreading them. That way it may be easier to know whether the config has loaded or not.

src/hooks/useSubscriptions.js
```diff
@@ -1,9 +1,33 @@
...
+
+  const configSubscription = useRef(null)
+
+  const onConfigHandler = useCallback(config => {
+    const transformedConfig = transformConfigData(config)
+    setConfig(transformedConfig)
+  }, [])
```
fabriziovigevani, line 16: Should we early return if there's no `config` ?

src/providers/AppState.js
```diff
@@ -41,6 +61,8 @@ function AppStateProvider({ children }) {
         stakeToken,
         totalSupply: totalSupply,
         vaultBalance,
+        totalStaked,
+        effectiveSupply,
```
fabriziovigevani, line 65: Also alphabetize this ones.

src/screens/ProposalDetail.js
```diff
@@ -295,18 +296,40 @@ const Amount = ({
...
-          icon={tokenIcon}
-        />
+        signalingProposal ? (
+          <div
+            css={`
+              display: flex;
+              align-items: center;
+            `}
+          >
+            <img src={signalingBadge} alt="" />
```
fabriziovigevani, line 312: Let's set the `height` and `width`

```suggestion
            <img src={signalingBadge} height="24" width="24"  alt="" />
```

src/screens/ProposalDetail.js
```diff
@@ -295,18 +296,40 @@ const Amount = ({
...
+            <span
+              css={`
+                margin-left: ${1 * GU}px;
+                ${textStyle('body2')};
+                font-weight: 300;
+              `}
+            >
+              {' '}
```
fabriziovigevani, line 320:
```suggestion

```

src/hooks/useProposals.js
```diff
@@ -123,7 +133,7 @@ export function useProposals() {
     maxRatio,
     proposals,
     stakesHistory,
-    totalSupply,
+    effectiveSupply,
```
fabriziovigevani, line 136: Let's order this one as well :)

rperez89:
> Awesome work!
> 
> Left some minor comments and suggestions.
> 
> Was also wondering if we should remove the `This proposal is requesting X Honey out of ...` message when it's a signaling proposal.
> 
> On a sidenote, i realized we must update as well wherever we are using `proposal.executed` and use instead `proposal.status`.

about the first one fiore suggested to let that message to emphasize, the second one good catch

src/hooks/useOrgHooks.js
rperez89: do you mind if we do this change after the release? don't want to touch a bunch of files right now? created https://github.com/1Hive/honey-pot/issues/59

src/app-logic.js
```diff
@@ -24,7 +25,10 @@ export default function useAppLogic() {
 
     return proposals.reduce(
       ({ myStakes }, proposal) => {
-        if (proposal.executed || !proposal.stakes) {
+        if (
+          proposal.status === PROPOSAL_STATUS_EXECUTED_STRING ||
```
fabriziovigevani, line 29: A proposal could also be cancelled.

```suggestion
          !proposal.status === PROPOSAL_STATUS_ACTIVE_STRING
```

rperez89: but we don't want to count the stakes on cancelled proposals?

PR #62, opened 2020-08-31
Deployment: gihub actions and vercel config

rperez89:
Needs the conviction voting app upgrade on xdai

### DIFF
--- a/.github/workflows/rinkeby.yml
+++ b/.github/workflows/rinkeby.yml
@@ -1,0 +1,31 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+name: rinkeby deployment
+
+on:
+  push:
+    branches:
+      -  master
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install node
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12
+    - name: yarn install
+      run: yarn install
+    - name: yarn lint
+      run: yarn run lint
+    - name: now
+      run: >
+        now
+        -A now-rinkeby.json
+        --confirm
+        --prod
+        --token=${{ secrets.ZEIT_TOKEN }}
+    env:
+      CI: true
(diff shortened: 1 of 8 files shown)


.github/workflows/rinkeby.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install node
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12
+    - name: yarn install
+      run: yarn install
+    - name: yarn lint
+      run: yarn run lint
```
fabriziovigevani, line 22: Maybe we could use something more descriptive ?

```suggestion
    - name: Install dependencies
      run: yarn install
    - name: Run linter
      run: yarn run lint
```

now-staging.json
```diff
@@ -0,0 +1,12 @@
+{
+    "version": 2,
+    "public": true,
+    "build": {
+      "env": {
+        "REACT_APP_CHAIN_ID": "4",
+        "REACT_APP_APP_NAME":"disputable-conviction-voting",
+        "REACT_APP_APP_RINKEBY_INSTANCE": "staging"
```
fabriziovigevani, line 8: What do you think of making it more general and not network sepcific? 
```suggestion
        "REACT_APP_APP_INSTANCE": "staging"
```

src/networks.js
```diff
@@ -1,5 +1,10 @@
 import { getNetworkType, isLocalOrUnknownNetwork } from './lib/web3-utils'
 import { getDefaultChain } from './local-settings'
+import env from './environment'
+
+const RINKEBY_HONEY_POT = '0xace2faa375ffc251fafc7360bc3defbc6184e5fe'
+const RINKEBY_STAGING_HONEY_POT = '0xeac000b64fc11a9ce6d885fe91fb4f9c2359cc21'
+const STAGING_ENV = env('STAGING')
```
fabriziovigevani, line 7: Maybe rename to `INSTANCE` ? Because `staging` is one of the possible values right? 
```suggestion
const INSTANCE = env('INSTANCE')
```

.github/workflows/rinkeby.yml
rperez89: ah, i think that like it is is ok because you have in the name the command that is going to be executed so is easier to see what is going on

PR #65, opened 2020-09-04
V2: Profile activity

fabriziovigevani:
Builds from #64 

Includes activity for supported proposals.
Does not include comments on proposals, will be added on a separate PR.
 
Also moves the `StakingTokens` and `Wallet` component to Profile page

Screen Shot 2020-09-10 at 11 58 34 AM

### DIFF
--- a/src/hooks/useStakes.js
+++ b/src/hooks/useStakes.js
@@ -1,0 +1,53 @@
+import { useMemo } from 'react'
+import { useAppState } from '../providers/AppState'
+import { useWallet } from '../providers/Wallet'
+
+import { addressesEqual } from '../lib/web3-utils'
+
+// TODO: See how to merge it with the implementation done at app-logic
+export function useMyStakes() {
+  const { account } = useWallet()
+  const { proposals } = useAppState()
+
+  return useMemo(
+    () =>
+      proposals.reduce((acc, proposal) => {
+        const myStake = proposal.stakes.find(
+          stake => addressesEqual(stake.entity, account) && stake.amount.gt(0)
+        )
+
+        if (!myStake) {
+          return acc
+        }
+
+        return [
+          ...acc,
+          {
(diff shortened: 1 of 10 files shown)


src/hooks/useStakes.js
```diff
@@ -0,0 +1,53 @@
...
+import { addressesEqual } from '../lib/web3-utils'
+
+// TODO: See how to merge it with the implementation done at app-logic
+export function useMyStakes() {
+  const { account } = useWallet()
+  const { proposals } = useAppState()
+
+  return useMemo(
+    () =>
+      proposals.reduce((acc, proposal) => {
```
rperez89, line 14: i think we already have this , you can do ` const {myStakes} = useAppLogic()`

src/providers/Profile.js
```diff
@@ -45,33 +47,22 @@ function ProfileProvider({ children }) {
...
 
-  const auth = useCallback(async () => {
-    try {
-      const box = await openBoxForAccount(account, ethereum)
-      setBox(box)
-    } catch (err) {
-      console.error(err)
-    }
-  }, [account, ethereum])
+    fetchPrivateData(box)
```
rperez89, line 64: let's add the cancel pattern

rperez89 (commented):
LGTM some minor comments

src/hooks/useStakes.js
fabriziovigevani: Ah yeah but the problem there is that we are also using other hooks that we don't want to use here, for example we are not really interested in the `useProposals` hook which will compute and calculate all conviction values. 

That's why i added a TODO there to see how to reuse it but not use the app-logic hooks that are not needed in this case.

rperez89: Agree on that! let's do it if possible

PR #66, opened 2020-09-08
V2: New header

fabriziovigevani:
build from #65 

Includes new [Header designs](https://www.figma.com/file/P06HTzSyb9hExJVpwaUPmA/Honey-Pot-2.0?node-id=46%3A961)

**Preview**

Screen Shot 2020-09-07 at 9 51 08 PM


src/components/Account/ScreenConnected.js
```diff
@@ -36,108 +36,131 @@ function AccountScreenConnected({ onClosePopover, wallet }) {
...
+          <div
+            css={`
+              display: flex;
+              align-items: center;
+              justify-content: flex-end;
+              width: 100%;
+            `}
+          >
+            <ButtonBase
+              onClick={() => copy(wallet.account)}
```
rperez89, line 113: could we create a handler for this?

src/components/Account/ScreenConnected.js
```diff
@@ -36,108 +36,131 @@ function AccountScreenConnected({ onClosePopover, wallet }) {
...
-          css={`
-            color: ${theme.contentSecondary};
-          `}
-        >
-          My profile
-        </Link>
-      </div>
-
       <Button
         onClick={() => wallet.deactivate()}
```
rperez89, line 165: same here

src/components/Account/ScreenConnected.js
```diff
@@ -36,108 +36,131 @@ function AccountScreenConnected({ onClosePopover, wallet }) {
...
           >
-            {`Connected to Ethereum ${walletNetworkName} Network`}
-          </span>
+            <IconCheck size="small" />
+            <span
+              css={`
+                margin-left: ${0.5 * GU}px;
+              `}
+            >
+              {`Connected to Ethereum ${walletNetworkName} Network`}
```
rperez89, line 158: i think that i already pointed this in a previous pr, but we might want to remove the Ethereum when we are on xdai

src/components/BalanceModule.js
```diff
@@ -0,0 +1,79 @@
...
+  const inactiveTokens = useMemo(() => {
+    if (!accountBalance.gte(0) || !myActiveTokens) {
+      return new BigNumber('0')
+    }
+    return accountBalance.minus(myActiveTokens)
+  }, [accountBalance, myActiveTokens])
+
+  const handleOnClick = useCallback(() => history.push('/profile'), [history]) // TODO: Send to profile/stake management section
+
+  const idlePct = inactiveTokens.eq('0')
```
rperez89, line 38: this makes more sense to be inactivePct right?

src/components/Header/HeaderModule.js
```diff
@@ -6,7 +6,7 @@ import {
   useTheme,
   useViewport,
 } from '@1hive/1hive-ui'
-function HeaderModule({ icon, content, onClick }) {
+function HeaderModule({ icon, content, hasPopover = true, onClick }) {
```
rperez89, line 9: lets alphabetize this

src/components/Header/HeaderPopover.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { GU, Popover, springs, textStyle, useTheme } from '@1hive/1hive-ui'
+import { GU, Popover, springs } from '@1hive/1hive-ui'
 import { Spring, animated } from 'react-spring/renderprops'
```
rperez89, line 4: also this import

src/lib/token-utils.js
```diff
@@ -40,7 +40,10 @@ const KNOWN_TOKENS_FALLBACK = new Map([
   ],
 ])
 
-const LOCAL_TOKEN_ICONS = new Map([['HNY', honeyIconSvg]])
+const LOCAL_TOKEN_ICONS = new Map([
+  ['HNY', honeyIconSvg],
+  ['CVTN', honeyIconSvg], // TODO: Remove
```
rperez89, line 45: what is this for?

rperez89 (commented):
awesome some minor comments!

src/components/Account/ScreenConnected.js
fabriziovigevani: Good catch, maybe we could remove `Ethereum` entirely and just leave the rest?

src/lib/token-utils.js
fabriziovigevani: Ah this is so we can see the token icon (which in this case is honey). Since we are mapping the icon from the token symbol, we wouldn't see it  when using  the Rinkeby env if we don't add the Rinkeby token.

src/components/Account/ScreenConnected.js
rperez89: yeah it makes sense

src/lib/token-utils.js
rperez89: ahhh cool

rperez89 (approved):
LGTM!

PR #67, opened 2020-09-08
V2 Identity badge: Add 3box profile name

fabriziovigevani:


src/components/IdentityBadge.js
```diff
@@ -1,13 +1,50 @@
...
+
+  useEffect(() => {
+    let cancelled = false
+    async function fetchProfile() {
+      if (addressCache.get(entity)) {
+        setProfileName(addressCache.get(entity))
+        return
+      }
+
+      const profile = await getProfileForAccount(entity)
```
rperez89, line 26: do we have some possible errors here?

rperez89 (commented):
just one question about the fetch :)

src/components/IdentityBadge.js
fabriziovigevani: It could error when fetching the public profile. I'll create an issue to work on general error handling so we can work on it after we have most of it merged!

rperez89: sounds good!

PR #68, opened 2020-09-08
Fixes: good first issues

PierrickGT:
Here is a PR fixing the following issue: 
- https://github.com/1Hive/honey-pot/issues/59

Waiting for an answer on the following issue, since I don't think it is a bug:
- https://github.com/1Hive/honey-pot/issues/60

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,5 +30,5 @@ yarn start:[network]
 
 ```
 REACT_APP_APP_NAME # Name of the app (if empty, defaults to conviction-voting)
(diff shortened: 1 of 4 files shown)


0xGabi:
Hey, @PierrickGT thanks a lot for your work. At the time the app was progressing super fast and we didn't manage to include the change in time. Now it got less relevant for the new direction we are heading towards. If you would like we are working on Gardens now: https://github.com/1Hive/gardens

And here is our [ZenHub board](https://app.zenhub.com/workspaces/gardens--6047c26dc05a1c000e95bcf3/board?filterLogic=any&repos=273752093,353881841,283226458,370756018).

PR #69, opened 2020-09-08
V2 Explore profiles

fabriziovigevani:
builds from #65 

Enables viewing other account's profiles. 

The url path for viewing other profiles is `/profile?account=<address>`

If no account is passed as query param it will default to the connected account's profile (if any).


src/screens/Profile.js
```diff
@@ -116,4 +142,9 @@ function AnimatedBackground({ height }) {
   )
 }
 
+function useQuery() {
```
rperez89, line 145: What about renaming it to useSearchParams? i needed to take a look to the code to understand what was that useQuery doing

rperez89 (commented):
Awesome! LGTM!

src/screens/Profile.js
fabriziovigevani: Yes sure!

PR #76, opened 2020-09-16
Subgraph

fabriziovigevani:
Honey pot subgraph for V2.

Currently deployed on Rinkeby: https://thegraph.com/explorer/subgraph/1hive/honey-pot-rinkeby?selected=playground
Queries this [new org](https://rinkeby.client.aragon.org/#/0x0381374d658b2c2e564e954219d9a3cfc6ae3fcb/) which is created from the honey-template. For reference see [this branch](https://github.com/1Hive/honey-template/tree/honey-karma) where we update the karma-template to use the updated `conviction-voting` app (`conviction-beta`) with the latest functionality.

This will allow to query mixed proposals:
 - Suggestions (a.k.a signaling proposals)
 - Decisions (a.k.a votes) 
  - Proposals (a.k.a proposals requesting for funds)

To achieve this, we created the `Proposal` entity which contains all mixed attributes, (added comments to separate which attributes corresponds to each). In the future we might want to asses if it makes sense to have the attributes in the respective entity and only share common ones, but we may have problems for querying specific data, so for now we can have them at the Proposal entity level.

TODO: 

- [ ] Add attributes for proposal ranking (Hot, Best, Top)


subgraph/manifest/data/rinkeby.json
```diff
@@ -0,0 +1,6 @@
+{
+  "network": "rinkeby",
+  "address": "0x0381374d658b2c2e564e954219d9a3cfc6ae3fcb",
+  "startBlock": 7207670,
```
rperez89, line 4: just to double check is the correct block number?

subgraph/schema.graphql
```diff
@@ -0,0 +1,112 @@
...
+
+type StakeHistory @entity {
+  id: ID!
+  type: StakeType!
+  entity: Bytes!
+  proposal: Proposal!
+  tokensStaked: BigInt!
+  totalTokensStaked: BigInt!
+  conviction: BigInt!
+  time: BigInt!      # Block at which was created
```
rperez89, line 86: i think that at some point we might want to change this naming hehe

subgraph/schema.graphql
```diff
@@ -0,0 +1,112 @@
...
+}
+
+enum ProposalStatus {
+  Active
+  Cancelled
+  Executed
+}
+
+enum ProposalType {
+  Suggestion # Signaling proposal
```
rperez89, line 104: opening a question to see what naming makes more sense , not sure if suggestion makes sense, on the decision one i remember at some point luke called them "administrative votes"

subgraph/src/DandelionVoting.ts
```diff
@@ -0,0 +1,52 @@
...
+  let proposal = getProposalEntity(event.address, event.params.voteId)
+
+  populateVotingDataFromEvent(proposal, event)
+  populateVotingDataFromContract(proposal, event.address, proposal.number)
+
+  proposal.type = PROPOSAL_TYPE_DECISION
+
+  proposal.save()
+}
+
```
rperez89, line 26: should we also have a handler for the config changes?

subgraph/src/helpers/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  if (config === null) {
+    config = new ConfigEntity(id)
+  }
+
+  return config
+}
+
+////// Proposal Entity //////
+export function getProposalEntityId(
```
rperez89, line 54: those two about proposal should be inside the ./conviction helpers right? and same with the config

rperez89 (commented):
awesome fabri! left some comments and questions

subgraph/manifest/data/rinkeby.json
fabriziovigevani: Yes, this is the one i deployed in Rinkeby for testing

subgraph/schema.graphql
fabriziovigevani: haha yeah maybe we could change it now actually (don't think will be a huge work updating the frontend), will do in the connector PR

subgraph/schema.graphql
fabriziovigevani: Hmm, i think using the same proposal type names as what we are going to show on the UI would make the most sense to me. https://www.figma.com/file/P06HTzSyb9hExJVpwaUPmA/Honey-Pot-2.0?node-id=62%3A1413

subgraph/src/DandelionVoting.ts
fabriziovigevani: Ah yeah good catch!

subgraph/src/helpers/index.ts
fabriziovigevani: No, because remember that the Proposal entity can be either a conviction proposal or a voting proposal.

Also the Config entity is the main entity which holds the conviction config and voting config (this two are handled inside each respective handler).

subgraph/src/DandelionVoting.ts
fabriziovigevani: Done in #77

PR #78, opened 2020-09-20
New app instance

fabriziovigevani:
Points to the new conviction voting app instance.

We need to redeploy the subgraph before re-deploying frontend.

Also fetches HNY price from Uniswap's v2 subgraph

### DIFF
--- a/src/components/AppLoader.js
+++ b/src/components/AppLoader.js
@@ -1,0 +1,6 @@
+import React from 'react'
+import { SyncIndicator } from '@1hive/1hive-ui'
+
+export default function AppLoader() {
+  return <SyncIndicator visible label="Loading app" />
+}
--- a/src/hooks/useUniswapHNYPrice.js
+++ b/src/hooks/useUniswapHNYPrice.js
@@ -1,0 +1,51 @@
+import { useEffect, useState } from 'react'
+
+import gql from 'graphql-tag'
+import { Client } from 'urql'
+
+const RETRY_EVERY = 3000
+
+const UNISWAP_URL = 'https://api.thegraph.com/subgraphs/name/1hive/uniswap-v2'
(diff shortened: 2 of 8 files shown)


src/components/AppLoader.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react'
+import { SyncIndicator } from '@1hive/1hive-ui'
+
+export default function AppLoader() {
+  return <SyncIndicator visible label="Loading app" />
```
rperez89, line 5: do you think we need this file just for this?

src/hooks/useUniswapHNYPrice.js
```diff
@@ -0,0 +1,51 @@
...
+
+  useEffect(() => {
+    let cancelled = false
+    let retryTimer
+    async function fetchPrice() {
+      try {
+        const result = await graphqlClient.query(HNY_PRICE_QUERY).toPromise()
+
+        const { pair } = result?.data
+        const hnyPrice = pair.token1Price
```
rperez89, line 32: should we do`pair?.token1Price`?

rperez89 (approved):
LGTM!

src/components/AppLoader.js
fabriziovigevani: Yes, this is just a placeholder for now. But Fiore is working on a custom loader so i though it was a good idea to already separate the component

src/hooks/useUniswapHNYPrice.js
fabriziovigevani: Ah actually we could return if `!result?.data`. If there;s data we can make sure that `token1Price` is available.

src/components/AppLoader.js
rperez89: cooool

PR #80, opened 2020-09-21
Fix BSOD due to supporters distribution component

sembrestels:
When we do not have space in the supporters distribution component and we need to have an "Others" line, the frontend crashes, and this patch is a fix to that issue.


befitsandpiper:
Looks good! I'd approve it, but I'm not seeing a way to.

fabriziovigevani (approved):
Thanks!

PR #83, opened 2020-09-21
Fix proposal fields

fabriziovigevani:
Proposal fields were not aligned with the values.

**Before:**
Screen Shot 2020-09-21 at 4 09 30 PM

**After**
Screen Shot 2020-09-21 at 4 09 22 PM


rperez89 (approved):
Awesome

PR #90, opened 2020-09-29
V1 fixes

fabriziovigevani:
Applies fixes done in v1.

### DIFF
--- a/src/components/BalanceModule.js
+++ b/src/components/BalanceModule.js
@@ -1,6 +1,6 @@
 import React, { useCallback, useMemo } from 'react'
 import { useHistory } from 'react-router-dom'
-import { textStyle, useTheme } from '@1hive/1hive-ui'
+import { formatTokenAmount, textStyle, useTheme } from '@1hive/1hive-ui'
 
 import HeaderModule from './Header/HeaderModule'
 import { useAppState } from '../providers/AppState'
@@ -8,7 +8,7 @@ import { useAccountStakes } from '../hooks/useStakes'
 import { useWallet } from '../providers/Wallet'
 
 import BigNumber from '../lib/bigNumber'
-import { formatTokenAmount, getTokenIconBySymbol } from '../lib/token-utils'
+import { getTokenIconBySymbol } from '../lib/token-utils'
 
 function BalanceModule() {
   const theme = useTheme()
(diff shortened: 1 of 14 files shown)


src/components/BalanceModule.js
```diff
@@ -67,7 +67,7 @@ function BalanceModule() {
                 color: ${theme.contentSecondary};
               `}
             >
-              ({inactivePct}% idle)
+              ({parseFloat(inactivePct).toFixed(2)}% idle)
```
rperez89, line 70: is this a js native function?

src/components/BalanceToken.js
```diff
@@ -39,7 +39,7 @@ const BalanceToken = ({ amount, symbol, color, size, icon }) => {
           icon || tokenIconUrl(tokenAddress, symbol, network && network.type)
         }
       />
-      {amount !== undefined ? splitAmount(amount.toFixed(3)) : ' - '}
+      {amount !== undefined ? splitAmount(amount, decimals) : ' - '}
```
rperez89, line 42: should we use null instead of undefined?

src/hooks/useFilterProposals.js
```diff
@@ -18,11 +18,11 @@ import { checkInitialLetters } from '../lib/search-utils'
 const NULL_FILTER_STATE = -1
 const STAKE_STATUS_FILTER_SUPPORTED = 1
 const STAKE_STATUS_FILTER_NOT_SUPPORTED = 2
-const EXECUTION_STATUS_FILTER_OPEN = 0
-const EXECUTION_STATUS_FILTER_ACCEPTED = 1
-const EXECUTION_STATUS_FILTER_CANCELLED = 2
-const TYPE_FILTER_FUNDING = 0
-const TYPE_FILTER_SIGNALING = 1
+const EXECUTION_STATUS_FILTER_OPEN = 1
```
rperez89, line 21: maybe we can alphabetize those constants

rperez89 (commented):
LGTM some minor comments!

src/components/BalanceModule.js
fabriziovigevani: Yes, its a float property function

src/hooks/useFilterProposals.js
fabriziovigevani: In this case i think its preferable the order to be by type of filter (given by the number) wdyt?

src/components/BalanceToken.js
fabriziovigevani: We could check for both values but i wouldn't drop the `undefined` check

rperez89 (approved):
LGTM

PR #91, opened 2020-09-29
Fixed Mobile UI wrapping

anthonyoliai:
I'm still kind of new to React and web development itself, but I wanted to help the honey community, so I started working on some simple issues. This should properly wrap the proposal text once the window's width gets too small. 

EDIT: I also noticed that the progress bar, is not wrapping properly. In Desktop mode, when the resolution gets small enough, the progress bar doesn't wrap properly. This PR also fixes that issue. Just a quick observation!

### DIFF
--- a/src/screens/Proposals.js
+++ b/src/screens/Proposals.js
@@ -106,6 +106,7 @@ const Proposals = React.memo(
                 display: flex;
                 align-items: center;
                 justify-content: space-between;
+                white-space:normal;
               `}
             >
               {account && (
@@ -189,6 +190,7 @@ const Proposals = React.memo(
                 id={proposal.id}
                 name={proposal.name}
                 selectProposal={handleSelectProposal}
+
               />,
             ]
             if (
@@ -229,7 +231,7 @@ const Proposals = React.memo(
                 />
               )
             }
-
+            
             return entriesElements
           }}
           tableRowHeight={14 * GU}
@@ -261,13 +263,16 @@ const IdAndTitle = ({ id, name, selectProposal }) => {
   }, [id, selectProposal])
 
   return (
-    <Link onClick={handleOnClick}>
+    <Link
+      onClick={handleOnClick}
+      css={`
+      white-space: normal;
+      text-align: left;
+    `}
+    >
       <span
         css={`
           color: ${theme.surfaceContentSecondary};
-          white-space:normal;
-          float:left;
-          text-align:left;
         `}
       >
         {name}
--- a/src/screens/Proposals.js
+++ b/src/screens/Proposals.js
@@ -265,6 +265,9 @@ const IdAndTitle = ({ id, name, selectProposal }) => {
       <span
         css={`
           color: ${theme.surfaceContentSecondary};
+          white-space:normal;
+          float:left;
+          text-align:left;
         `}
       >
         {name}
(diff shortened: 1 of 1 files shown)


sembrestels:
Thanks @anthonyoliai for taking the time for doing this!

src/screens/Proposals.js
```diff
@@ -265,6 +265,9 @@ const IdAndTitle = ({ id, name, selectProposal }) => {
       <span
         css={`
           color: ${theme.surfaceContentSecondary};
+          white-space:normal;
+          float:left;
+          text-align:left;
```
fabriziovigevani, line 270: What do you think if we apply this styles to the parent `Link` component? In that case we can drop the `float` prop.

fabriziovigevani (commented):
Thanks @anthonyoliai for this! Left one suggestion.

Also you mentioned you did a fix on the progress bar that you might have forgotten to push 😄

src/screens/Proposals.js
anthonyoliai: Hi , yes sorry i'll try to get it figured out. Sorry still new to using github 😅

fabriziovigevani: haha no to worry!  Thank you

anthonyoliai: Haha alright. I'm still learning as i'm going, as far as I can see you recommended me to apply the style to the parent Link component. I've however been having a hard time finding this as I am new to the React infrastructure. How I see it is that the modified Link component is part of `const IdAndTitle` , which is in turn part of a large DataView `Proposals.js line 185-235 ` I've been trying a couple of things out applying style changes for different components to no success.  I feel like I'm not looking at the right place, tips would be very appreciated as all of this is a learning process for me :)

fabriziovigevani: Ah sorry, maybe i should have been more specific! I was referring to do something like this:

```
    <Link
      onClick={handleOnClick}
      css={`
        white-space: normal;
        text-align: left;
      `}
    >
      <span
        css={`
        `}
      >
        {name}
      </span>
    </Link>
```

anthonyoliai: Done! Sorry for the wait. Also, by taking this styling approaches the progress bar is automatically fixed aswell, so for that nothing else had to be pushed. To recreate: On Desktop mode when resizing the screen resolution of the browser, the progress bar will wrap properly.

src/screens/Proposals.js
```diff
@@ -261,7 +261,13 @@ const IdAndTitle = ({ id, name, selectProposal }) => {
   }, [id, selectProposal])
 
   return (
-    <Link onClick={handleOnClick}>
+    <Link
+      onClick={handleOnClick}
+      css={`
+      white-space: normal;
+      text-align: left;
+    `}
```
fabriziovigevani, line 269: Ah, could we fix the formatting? 

```suggestion
      css={`
        white-space: normal;
        text-align: left;
      `}
```

anthonyoliai: Done! Sorry been a long day.

rperez89: ah this formatting stuff might be fixed when you save the changes to the file thanks to the .eslint configuration, if that is not working to you let us know

rperez89 (commented):
nice!!

fabriziovigevani (approved):
🙌  Thanks @anthonyoliai !

PR #93, opened 2020-09-30
Search function improved for 1hive.org

anthonyoliai:
The search function on 1hive.org was improved. This was in order to fix a bug that caused a client-side crash when entering special regex characters such as '*'.  Changes were tested, no conflicting errors.


sembrestels:
Could you add the `\` char too in the characters denylist, please?

anthonyoliai:
Added!

fabriziovigevani (approved):
LGTM! Thanks!

PR #94, opened 2020-10-01
AccountModule: fix 3box onboarding modal overlapping 

rperez89:
Screen Shot 2020-10-01 at 12 19 59 PM

### DIFF
--- a/src/lib/profile.js
+++ b/src/lib/profile.js
@@ -49,6 +49,7 @@ function parseProfileData(profile, verifiedAccounts) {
 
   return {
     ...profile,
+    profileExists: Boolean(Object.keys(profile).length),
     coverPhoto,
     image,
     verifiedAccounts: parsedVerifiedAccounts,
--- a/src/lib/profile.js
+++ b/src/lib/profile.js
@@ -49,7 +49,6 @@ function parseProfileData(profile, verifiedAccounts) {
 
   return {
     ...profile,
-    proof_did: profile?.proof_did,
     coverPhoto,
     image,
     verifiedAccounts: parsedVerifiedAccounts,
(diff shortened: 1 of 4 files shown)


src/lib/profile.js
```diff
@@ -49,6 +49,7 @@ function parseProfileData(profile, verifiedAccounts) {
 
   return {
     ...profile,
+    profileExists: Boolean(Object.keys(profile).length),
```
fabriziovigevani, line 52: Maybe we could early return in `getProfileForAccount` if profile doesn't exist and just return true here ?
Also lets order it.

src/providers/Profile.js
```diff
@@ -38,6 +39,11 @@ function ProfileProvider({ children }) {
         setBox(box)
       }
     } catch (err) {
+      setProfile(profile => ({
+        ...profile,
+        onboardingSkipped: true,
```
fabriziovigevani, line 44: Would rename to `confirmationFailed` because it's not something you can skip, if you don't confirm the modal you won't be authenticated.

src/providers/Profile.js
```diff
@@ -60,6 +70,7 @@ function ProfileProvider({ children }) {
 
   useEffect(() => {
     setProfile(null)
+    setLoadingProfile(true)
```
fabriziovigevani, line 73: Should we move this after the if statement below?

Or we could actually move it to `fetchAccountProfile`

fabriziovigevani (commented):
Cool, left some comments!

We are still having the issue when the profile is already created and is the first time the user is connecting their profile with the app.

src/lib/profile.js
rperez89: will do this in another pr when your fixes get merged

Issue #92, opened 2020-09-30
Search bar in mobile version


sembrestels:
As @anthonyoliai posted recently in discord:
> I noticed that mobile users cannot search for proposals on the mobile version of the 1hive.org website. There is no search bar present. I propose that we do add this functionality as it may be useful for the user.

PR #95, opened 2020-10-02
Mobile search

anthonyoliai:
As proposed in #92: A simple mobile search, one small bug is that the margins are still in place when the screen gets really small so it's not perfectly aligned. This is due to the nature of flex-wrap. Right now the dropdowns work, and the search bar is a popover. Can make it so that it's already an extended bar if suggested. I just thought it would look cleaner this way.

### DIFF
--- a/src/components/FilterBar/TextFilter.js
+++ b/src/components/FilterBar/TextFilter.js
@@ -29,33 +29,14 @@ const TextFilter = React.memo(
           margin-left: ${1.5 * GU}px;
         `}
       >
-        {!compactMode ? (
-          <SearchInput
-            value={textFilter}
-            onChange={updateTextFilter}
-            placeholder={placeholder}
-            css={`
+        <SearchInput
+          value={textFilter}
+          onChange={updateTextFilter}
+          placeholder={placeholder}
+          css={`
               width: ${32 * GU}px;
             `}
-          />
-        ) : (
-          <React.Fragment>
-            <Button
-              display="icon"
-              icon={<IconSearch />}
-              ref={openerRef}
-              label="Search Proposal"
-              onClick={onClick}
-            />
-            <TextFilterPopover
-              textFilter={textFilter}
-              updateTextFilter={updateTextFilter}
-              visible={visible}
-              setVisible={setVisible}
-              opener={openerRef.current}
-            />
-          </React.Fragment>
-        )}
+        />
       </div>
     )
   }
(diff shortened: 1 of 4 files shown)


rperez89 (changes requested):
Hey man! thxs! could you please share some screenshots to see how it looks likes? we should ask @fioreb about a desing for this, also cc @fabriziovigevani i remember that there were some design guideline or something about this because in any aragon app we have search bar on mobile view

fioreb:
Thanks!

Filters and search bar were designed by the Aragon1 team but the implementation was left for later, for the moment in mobile devices we don't have filters of any type because of that.

Since we are going to change for a Feed in Honey Pot version 2, I add the filters to the designs but we don't have a search bar. Here you can check the designs: https://www.figma.com/file/P06HTzSyb9hExJVpwaUPmA/Honey-Pot-2.0?node-id=461%3A7943

I think that if we are going to add the search bar now, we can put it always displayed as we have it on desktop

anthonyoliai:
Sounds great! Thanks for the review and replies. I've mostly just reused the same components for the FilterBar that were already in place, to prevent any design inconsistencies. 
2aad6c8c9df0711221b6ef004e58cd41
Perhaps it's better to just remove the popover in the mobile version and have it already be an expanded search bar.

fioreb:
Great, thanks for that!

Yes, I think that have it already as an expanded search bar would be nice!

anthonyoliai:
Alright! I did what I could :) (Still learning how React works..). What I did is use a negative margin-left on the parent div to fix the flex-wrapping problem, which caused the dropdown filters not to be aligned when wrapping due to a small width. Two examples given: normal mobile width, very small mobile width:
screenshot1
screenshot2
Search functionality works :)

src/components/FilterBar/TextFilter.js
```diff
@@ -1,61 +1,26 @@
...
-    textFilter,
-    updateTextFilter,
-    placeholder = '',
-    openerRef,
-    visible,
-    setVisible,
-    onClick,
-  }) => {
+  ({ textFilter, updateTextFilter, placeholder = '' }) => {
     const { layoutName } = useLayout()
```
fabriziovigevani, line 8: I think we should remove this one :)

src/components/FilterBar/TextFilter.js
```diff
@@ -1,61 +1,26 @@
...
 
 const TextFilter = React.memo(
-  ({
-    textFilter,
-    updateTextFilter,
-    placeholder = '',
-    openerRef,
-    visible,
-    setVisible,
-    onClick,
```
fabriziovigevani, line 21: Since we are removing this props, we should also remove them from the declared `PropTypes` below

fabriziovigevani (commented):
Nicely done @anthonyoliai !

Left some comments

Was also wondering if we should use some bottom margin between filters when they are stacked?

anthonyoliai:
Applied the changes, search filters indeed look nicing with bottom margin. Also cleaned the overall code! Please have a look :) Let me know if more changes are necessary.

fabriziovigevani (commented):
Looks good!

We missed to add the same bottom margin to the `DropwdownFilter` which is displayed conditionally from the proposal status filter (Removed or Closed).


Screen Shot 2020-10-07 at 11 02 35 AM

Lets add that and LGTM!

fabriziovigevani (approved):
LGTM! Thanks!

PR #97, opened 2020-10-03
Network Error Modal

famole:
I'm showing the error message based on the conviction voting exists, on the other hand the isLoading state is based using the same condition.
https://github.com/famole/honey-pot/blob/c28001072e34e248213c1bf81b0fa5f0156aea3f/src/providers/AppState.js#L47
I'm wondering how we can fix this cc: @rperez89 @fabriziovigevani 

Screen Shot 2020-10-03 at 4 01 57 PM


rperez89:
@fabriziovigevani was wondering since `connect` doesn't provide us with any error connection or timeout if might make sense to here https://github.com/famole/honey-pot/blob/c28001072e34e248213c1bf81b0fa5f0156aea3f/src/providers/AppState.js#L47 add a timeout like 15 seconds or something like that and change the loading state to false if so

rperez89:
@famole we just figured out a way to do it with fabri, let me know once you are available and we can do it

src/App.js
```diff
@@ -10,6 +10,7 @@ import Wallet from './components/Wallet'
 import useAppLogic from './app-logic'
 import { useWallet } from './providers/Wallet'
 import useSelectedProposal from './hooks/useSelectedProposal'
+import NetworkErrorModal from './components/NetworkErrorModal'
```
fabriziovigevani, line 13: Could we move this to above where we have the component imports? 

We generally order the imports in the following way:

- External library imports
- Component imports 
- Hook imports
- local utilities, library imports

And we also order them in alphabetical order 😄

src/app-logic.js
```diff
@@ -11,7 +11,7 @@ import { PROPOSAL_STATUS_ACTIVE_STRING } from './constants'
 export default function useAppLogic() {
   const { account } = useWallet()
 
-  const { isLoading, stakeToken, totalStaked } = useAppState()
+  const { isLoading, stakeToken, totalStaked, errorFetchingApp, errorFetchingOrg } = useAppState()
```
fabriziovigevani, line 14: What do you think if we only expose an `errors` variable from the AppStatePRovider which combines the logic of both errors?

src/providers/AppState.js
```diff
@@ -44,7 +45,7 @@ function AppStateProvider({ children }) {
   }, [totalSupply, totalStaked, minThresholdStakePercentage])
 
   const balancesLoading = vaultBalance.eq(-1) || totalSupply.eq(-1)
-  const appLoading = !convictionVoting || balancesLoading || !effectiveSupply
+  const appLoading = errorFetchingApp || errorFetchingOrg ? false : !convictionVoting || balancesLoading || !effectiveSupply
```
fabriziovigevani, line 48: As commented above, we could refactor this a little bit to be:

```suggestion
  const errors =  errorFetchingApp || errorFetchingOrg
  const appLoading =  !errors && !convictionVoting || balancesLoading || !effectiveSupply
```

src/providers/AppState.js
```diff
@@ -53,6 +54,7 @@ function AppStateProvider({ children }) {
         accountBalance: balance,
         convictionVoting,
         effectiveSupply,
+        errorFetchingOrg,
```
fabriziovigevani, line 57: And here pass errors instead.
```suggestion
       errors,
```

src/components/NetworkErrorModal.js
```diff
@@ -0,0 +1,38 @@
...
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+            text-align:center;
+            `}
+        >
+            <img src={flowerError} alt="" height="88" width="71"/>
+            <h3
+            css={`
+                ${textStyle('title2')}
```
fabriziovigevani, line 21:
```suggestion
                ${textStyle('title2')};
```

src/components/NetworkErrorModal.js
```diff
@@ -0,0 +1,38 @@
...
+                margin-top: 24px;
+                margin-bottom: 8px;
+            `}>
+            Something went wrong
+            </h3>
+            <h4
+                css={`
+                    ${textStyle("body3")}
```
fabriziovigevani, line 29:
```suggestion
                    ${textStyle("body3")};
```

fabriziovigevani (commented):
Awesome work Fabi!

Left some comments, let me know what do you think.

Also i noticed the code formatting is missing. We use `prettier` in our projects which basically auto formats your code after saving new changes, where the `.eslintrc` is the formatting configuration file. 
If you use vscode let us know and we can help you set it up!

src/App.js
famole: Hey Fabri, 

Thank you for the explanation!!, I'll change these things.
I've just added prettier to my vsCode, will be checking with Rodri in case some configuration missing.

src/app-logic.js
```diff
@@ -56,14 +56,15 @@ export default function useAppLogic() {
 
   const actions = useActions(proposalPanel.requestClose)
 
+  console.log("ERRORS: ", fetchingErrors);
```
fabriziovigevani, line 59:
```suggestion
```

fabriziovigevani (approved):
Awesome @famole, lets remove the console.log and merge 🙌

fabriziovigevani (approved):
Thanks @famole !

PR #101, opened 2020-10-07
Simple icon redirect

anthonyoliai:
Had someone comment , especially in mobile that they wanted to be able to get redirected by tapping on the 1Hive icon. Quickly made a redirect, not sure if there is a cleaner way of doing this :)

### DIFF
--- a/src/components/Header.js
+++ b/src/components/Header.js
@@ -8,7 +8,13 @@ import headerBackgroundSvg from '../assets/header-background.svg'
 import logoSvg from '../assets/logotext.svg'
 
 function Header({ compact }) {
-  const BeeIcon = <img src={beeSvg} height={compact ? 40 : 60} alt="" />
+  const BeeIcon = (
+    <a href="https://1hive.org">
+      {' '}
+      <img src={beeSvg} height={compact ? 40 : 60} alt="" />
+    </a>
+  )
+
   const headerItemsWidth = compact ? 'auto' : 25 * GU
 
   return (
--- a/src/components/Header.js
+++ b/src/components/Header.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { GU } from '@1hive/1hive-ui'
+import { GU, Link } from '@1hive/1hive-ui'
 import AccountModule from './Account/AccountModule'
 import Layout from './Layout'
 
@@ -9,9 +9,9 @@ import logoSvg from '../assets/logotext.svg'
 
 function Header({ compact }) {
   const BeeIcon = (
-    <a href="https://1hive.org">
+    <Link href="/" external={false}>
       <img src={beeSvg} height={compact ? 40 : 60} alt="" />
-    </a>
+    </Link>
   )
 
   const headerItemsWidth = compact ? 'auto' : 25 * GU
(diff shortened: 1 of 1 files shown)


src/components/Header.js
```diff
@@ -8,7 +8,13 @@ import headerBackgroundSvg from '../assets/header-background.svg'
 import logoSvg from '../assets/logotext.svg'
 
 function Header({ compact }) {
-  const BeeIcon = <img src={beeSvg} height={compact ? 40 : 60} alt="" />
+  const BeeIcon = (
+    <a href="https://1hive.org">
+      {' '}
```
rperez89, line 13: i think we can remove this space right?

rperez89 (commented):
Thxs man, now that you are getting more involved i think that might be good if you can reach me or @fabriziovigevani before submitting a PR for a feature requested in the community because we are launching the v2 soon and you might end working on stuffs that might not be needed! just to coordinate!

anthonyoliai:
> Thxs man, now that you are getting more involved i think that might be good if you can reach me or @fabriziovigevani before submitting a PR for a feature requested in the community because we are launching the v2 soon and you might end working on stuffs that might not be needed! just to coordinate!

Alright! Very sorry about that, I will coordinate more :) sorry still getting used to this work setting 😅

anthonyoliai:
I had a talk earlier with @sembrestels today regarding honeypot v1, as the changes in v1 could be handy to v2 too. And we've come to the conclusions that there are some simple quality of life problems such as this that can be easily fixable. Hence the follow up PR, hope you can have a look.

src/components/Header.js
```diff
@@ -8,7 +8,12 @@ import headerBackgroundSvg from '../assets/header-background.svg'
 import logoSvg from '../assets/logotext.svg'
 
 function Header({ compact }) {
-  const BeeIcon = <img src={beeSvg} height={compact ? 40 : 60} alt="" />
+  const BeeIcon = (
+    <a href="https://1hive.org">
```
fabriziovigevani, line 12: We can use the [`Link`](https://ui.aragon.org/link/) component. from `@1hive-ui` in this case, we can also use relative path.

```suggestion
    <Link href="/" external={false}>
```

fabriziovigevani:
I think as long as they are adding small features that bring value to the current version it's fine. And i agree they can also be used for V2. But yeah when is a feature or change more involved then let's discuss if it's worth doing it for this version or v2 :)

src/components/Header.js
anthonyoliai: That's a much cleaner way of doing it! Shows how inexperienced I am with React as of now haha. Putting up a new PR, thanks!

src/components/Header.js
```diff
@@ -9,9 +9,9 @@ import logoSvg from '../assets/logotext.svg'
 
 function Header({ compact }) {
   const BeeIcon = (
-    <a href="https://1hive.org">
+    <Link href="/" external={false}>
```
fabriziovigevani, line 12: Ah sorry, forgot we are using a hash router, lets add the `#`

```suggestion
    <Link href="/#" external={false}>
```

PR #102, opened 2020-10-09
BrightId verification

rperez89:
Screen Shot 2020-10-09 at 4 16 57 PM

Screen Shot 2020-10-09 at 4 17 31 PM
Screen Shot 2020-10-09 at 4 17 37 PM
Screen Shot 2020-10-09 at 4 17 44 PM


fabriziovigevani (commented):
Awesome Ro!

A couple of comments regarding the brighID fetch.

What do you think if instead of fetching the BrightId info from the Profile provider, we do it from the `BrightIDStatus` component? 
The reason i suggest this is that we are doing two polls from a context provider where the data we are fetching isn not going to be used across all the app. 
Also the profile object would be a new object on every poll in this way.

If we do this, also we may want to render the BrightIdModal inside the BrightIdStatus component so we encapsulate all that logic.

fabriziovigevani (approved):
LGTM Ro.

Let's check with BrightId if they can give 1hive.org access before merging.

PR #105, opened 2020-10-11
Add proposal: Require forum post link

fabriziovigevani:
Requires a link to the corresponding 1hive forum post when creating a proposal.

**Preview**
Screen Shot 2020-10-10 at 10 09 25 PM

The `from here` link is a link to [this template](https://forum.1hive.org/t/creating-better-proposals/200/2)

### DIFF
--- a/src/components/AddProposalPanel.js
+++ b/src/components/AddProposalPanel.js
@@ -6,6 +6,7 @@ import {
   GU,
   Info,
   isAddress,
+  Link,
   MEDIUM_RADIUS,
   TextInput,
   useTheme,
@@ -18,6 +19,7 @@ import { formatTokenAmount } from '../lib/token-utils'
 import { calculateThreshold, getMaxConviction } from '../lib/conviction'
 
 import { ZERO_ADDR } from '../constants'
+const FORUM_POST_REGEX = /https:\/\/forum.1hive.org\/t\/.*?\/([0-9]+)/
 
 const NULL_PROPOSAL_TYPE = -1
 const FUNDING_PROPOSAL = 1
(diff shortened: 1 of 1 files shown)


src/components/AddProposalPanel.js
```diff
@@ -153,11 +155,13 @@ const AddProposalPanel = React.memo(({ onSubmit }) => {
       if (beneficiary && !isAddress(beneficiary)) {
         errors.push('Beneficiary is not a valid ethereum address')
       }
+    }
 
-      return errors
+    if (link && !FORUM_POST_REGEX.test(link)) {
```
rperez89, line 160: wondering if at some point we might want to do a check on the smart contract to avoid direct calls without it

rperez89 (approved):
LGTM!

0xGabi (approved):

src/components/AddProposalPanel.js
sembrestels: I don't think it will be a problem since we are recommending to use some proposal guidelines.

sembrestels:
The [commit suggestions](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request#applying-a-suggested-change) are not available in this PR, so I just added another commit that uses just this link (note that the proposals category already has the prefilled template):

https://forum.1hive.org/new-topic?category=proposals 

I deleted the second sentence because I think it is better to have just one call for action.

sembrestels (approved):
I committed some changes in the info box, but the rest looks good to me.

src/components/AddProposalPanel.js
rperez89: but what prevents to call the contract regardless there are guidelines?

fabriziovigevani: I don't really see a reason to go trough that hurdle just to bypass the link requirement.

PR #106, opened 2020-10-12
Refactor proposal slider to fix rounding issue

sembrestels:
I refactored the component to use big numbers, so it's much more precise now.

### DIFF
--- a/src/components/Metrics.js
+++ b/src/components/Metrics.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Box, GU, textStyle, Link, useLayout, useTheme } from '@1hive/1hive-ui'
 
-import { formatTokenAmount } from '../lib/token-utils'
+import { formatTokenAmount, formatDecimals } from '../lib/token-utils'
 import honeySvg from '../assets/honey.svg'
 import { useUniswapHnyPrice } from '../hooks/useUniswapHNYPrice'
 
@@ -128,7 +128,11 @@ function TokenPrice({ token }) {
 
   return (
     <div>
-      <Metric label="Honey price" value={`$${price}`} color={theme.green} />
+      <Metric
+        label="Honey price"
+        value={`$${formatDecimals(price, 2)}`}
+        color={theme.green}
+      />
       <Link
         href={`https://honeyswap.org/#/swap?inputCurrency=${token.id}`}
         external
--- a/src/components/ProposalActions.js
+++ b/src/components/ProposalActions.js
@@ -38,20 +38,13 @@ function ProposalActions({
     [myStake.amount, accountBalance, totalStaked]
   )
 
-  const roundSlider = bigNum =>
-    bigNum
-      .shiftedBy(-stakeToken.decimals)
-      .decimalPlaces(Math.min(MAX_INPUT_DECIMAL_BASE, stakeToken.decimals))
-      .toString()
+  const [
+    inputValue, // tokens amount formatted as a string
+    amount, // tokens amount formatted as a big number
+    handleAmountChange,
+    handleSliderChange,
+  ] = useAmount(myStake, stakeToken, maxAvailable)
 
-  const [inputValue, setInputValue] = useState(
-    myStake?.amount &&
-      fromDecimals(myStake.amount.toString(), stakeToken.decimals)
-  )
-  const amount = BigNumber.minimum(
-    new BigNumber(toDecimals(inputValue, stakeToken.decimals)),
-    maxAvailable
-  )
   const didIStake = myStake?.amount.gt(0)
 
   const mode = useMemo(() => {
@@ -87,7 +80,7 @@ function ProposalActions({
         .integerValue()
         .toString(10)
     )
-  }, [id, amount, myStake.amount, onStakeToProposal, onWithdrawFromProposal])
+  }, [amount, id, myStake.amount, onStakeToProposal, onWithdrawFromProposal])
 
   const buttonProps = useMemo(() => {
     if (mode === 'execute') {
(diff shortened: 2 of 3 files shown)


rperez89 (commented):
cool! just a comment about the implementation i think we should keep the `useAmount` hook and left there encapsulated all the slider - amounts logic since is more clean, and add the stuffs that you did there instead of the main component

src/components/Metrics.js
```diff
@@ -128,7 +128,11 @@ function TokenPrice({ token }) {
 
   return (
     <div>
-      <Metric label="Honey price" value={`$${price}`} color={theme.green} />
+      <Metric
+        label="Honey price"
+        value={`$${formatDecimals(price, 2)}`}
```
fabriziovigevani, line 133: What's the reason for adding this? Note that we are already fixing it to 2 decimal positions https://github.com/1Hive/honey-pot/blob/8d9c08c9d60ddbe05f1feaf9fdf992ae7d8af65b/src/hooks/useUniswapHNYPrice.js#L39

src/components/ProposalActions.js
```diff
@@ -123,13 +114,13 @@ function ProposalActions({
...
+    onRequestSupportProposal,
     accountBalance,
     handleExecute,
+    amount,
     handleChangeSupport,
-    inputValue,
-    mode,
-    myStakeAmountFormatted,
-    onRequestSupportProposal,
+    myStake.amount,
```
fabriziovigevani, line 123: Could we alphabetize them back again?

src/components/ProposalActions.js
```diff
@@ -80,22 +69,25 @@ function ProposalActions({
...
-    onWithdrawFromProposal,
-  ])
+    onStakeToProposal(
+      id,
+      amount
+        .minus(myStake.amount)
+        .integerValue()
+        .toString(10)
+    )
+  }, [id, amount, myStake.amount, onStakeToProposal, onWithdrawFromProposal])
```
fabriziovigevani, line 90:
```suggestion
  }, [amount, id, myStake.amount, onStakeToProposal, onWithdrawFromProposal])
```

src/components/ProposalActions.js
```diff
@@ -143,20 +134,26 @@ function ProposalActions({
             `}
           >
             <Slider
-              value={progress}
-              onUpdate={setProgress}
+              value={
+                maxAvailable.gt(0) ? amount.div(maxAvailable).toNumber() : 0
+              }
+              onUpdate={newProgress =>
```
fabriziovigevani, line 140: Let's memoize this functions, with `useCallback`, otherwise they'll be re created on every re-render.

src/components/ProposalActions.js
```diff
@@ -143,20 +134,26 @@ function ProposalActions({
...
                 padding-left: 0;
                 width: 100%;
               `}
             />
             <TextInput
               value={inputValue}
-              onChange={setAmount}
               type="number"
-              max={maxAvailable}
+              onChange={event => setInputValue(event.target.value)}
```
fabriziovigevani, line 153: Same here

src/components/ProposalActions.js
```diff
@@ -34,35 +33,25 @@ function ProposalActions({
...
-
-  const formattedMaxAvailableAmount = useMemo(() => {
-    if (!stakeToken) {
-      return '0'
-    }
-    return formatTokenAmount(nonStakedTokens, stakeToken.decimals)
-  }, [stakeToken, nonStakedTokens])
-
-  const rounding = Math.min(MAX_INPUT_DECIMAL_BASE, stakeToken.decimals)
+  const roundSlider = bigNum =>
```
fabriziovigevani, line 41: Let's also memoize this one.

fabriziovigevani (commented):
Awesome improvement @sembrestels 

Left some comments, i also agree with @rperez89 that we should re-encapsulate the `amount` logic into a hook, so we have a cleaner code.

src/components/Metrics.js
sembrestels: To keep the a coherent format with the rest of values in dollars. The converted prices use the `formatDecimals` function, that puts comma separators. I realized that when the price of HNY is >$1,000, it doesn't have that comma separator.

src/components/ProposalActions.js
sembrestels: Done.

src/components/ProposalActions.js
sembrestels: How can this function be putted in a callback? It doesn't depend on any other parameter.

src/components/ProposalActions.js
sembrestels: Done.

src/components/ProposalActions.js
sembrestels: Done.

src/components/ProposalActions.js
sembrestels: Done.

sembrestels:
I followed your comments but I'm not sure if I'm using `useCallback` properly (I'm learning something today!!). Can you check it out again please?

src/components/ProposalActions.js
fabriziovigevani: When it doesn't depend on any other parameter, it will just be the same function on every render.
In this cases we leave the array of dependencies empty :)

src/components/Metrics.js
fabriziovigevani: Ah good catch!

fabriziovigevani:
@sembrestels Just commented! Apart from that missing `useCallback` LGTM

PR #108, opened 2020-10-12
App reestructure

fabriziovigevani:
- Uses Honeypot connector
- Updates how data is handled because of the new subgraph schema, mostly on `useOrgHooks` and `AppStateProvider`
- Rearranges a bit the main views components and reorders them logically by screen.
- Implements basic Proposal list (only displays proposal id) 
- Implements proposal filters logic (no UI)

On the next PR will push the proposals filters UI and proposal card UI.


src/hooks/useProposalFilters.js
```diff
@@ -0,0 +1,117 @@
...
+    setTypeFilter(NULL_FILTER_STATE)
+  }, [])
+
+  useEffect(() => {
+    if (!account) {
+      setSupportFilter(NULL_FILTER_STATE)
+    }
+  }, [account])
+
+  const isActive =
```
rperez89, line 58: this active is to know if the proposal is still active? if so, shouldn't we just check for the status?

rperez89 (commented):
A W E S O M E!

src/components/MainView.js
```diff
@@ -1,9 +1,10 @@
...
-import { useViewport } from '@1hive/1hive-ui'
+import { GU, Root, useViewport } from '@1hive/1hive-ui'
 
 import Footer from './Footer'
 import Header from './Header/Header'
 import Layout from './Layout'
+import ScrollView from './ScrollView'
 
 function MainView({ children }) {
   const { below } = useViewport()
```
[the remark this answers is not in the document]
fabriziovigevani: In this case we cannot use it because MainView is not inside  a `Layout` component

src/hooks/useEffectiveSupply.js
```diff
@@ -0,0 +1,20 @@
+import { useMemo } from 'react'
+import { STAKE_PCT_BASE } from '../constants'
+
+export default function useEffectiveSupply(totalSupply, config) {
```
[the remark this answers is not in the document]
fabriziovigevani: Ah yeah, this was taken from master yes

src/hooks/useProposalFilters.js
fabriziovigevani: Ah no, this is to indicate that the filters are active (meaning they are not the default ones). Do you think the naming is confusing?

rperez89: ah maybe yes, we can rename it to filtersActive or something like that maybe

rperez89:
Going to merge and we can address the isActive name change in another one

PR #113, opened 2020-10-14
V2: New feed fixes

fabriziovigevani:
Implements all that was done in `v2-feed` branch but not included in the previously split PRs.

Also fixes the projected conviction not showing previously.


src/components/Account/ScreenConnected.js
```diff
@@ -41,31 +42,28 @@ function AccountScreenConnected({ onClosePopover, wallet }) {
...
-          Account
-        </h4>
-        <Link
-          onClick={goToProfile}
-          external={false}
-          css={`
-            color: ${theme.content};
+            padding-bottom: ${2 * GU}px;
+            border-bottom: 1px solid ${theme.border};
+
```
rperez89, line 57: lets remove this blank space

src/components/BalanceToken.js
```diff
@@ -47,7 +47,7 @@ const BalanceToken = ({ amount, color, decimals, icon, size, symbol }) => {
   )
 }
 
-const TokenIcon = styled.img.attrs({ alt: '', width: '20', height: '20' })`
+const TokenIcon = styled.img.attrs({ alt: '', width: '24', height: '24' })`
```
rperez89, line 50: are we using this image like this for some reason?

src/components/Profile/Activity.js
```diff
@@ -1,11 +1,35 @@
...
+import { dateFormat } from '../../utils/date-utils'
+
 function Activity({ account }) {
   const theme = useTheme()
 
-  const myStakeHistory = []
+  const { honeypot } = useAppState()
+  const supporter = useSupporterSubscription(honeypot, account)
+
+  const dedupedStakes = useMemo(() => {
```
rperez89, line 17: what do you mean by deduped stakes here?

src/hooks/useAccountTotalStaked.js
```diff
@@ -1,31 +1,16 @@
...
-          )
-
-          if (!myStake) {
-            return acc
-          }
-
-          return acc.plus(myStake.amount)
-        }, new BigNumber(0)),
-    [proposals, account]
+      stakes.reduce((acc, { amount }) => {
```
rperez89, line 11: just to double check are we returning only the active ones on the useAccountStakes right?

src/utils/date-utils.js
```diff
@@ -7,6 +7,7 @@ import advancedFormat from 'dayjs/plugin/advancedFormat'
 const KNOWN_FORMATS = {
   onlyDate: 'DD/MM/YYYY',
   iso: 'YYYY-MM-DD',
+  custom: 'DD MMMM HH:mm',
```
rperez89, line 10: hmm 4 M? which format is that?

rperez89 (approved):
LGTM  only some questions

src/components/BalanceToken.js
fabriziovigevani: Think this was forked from the original Aragon client Conviction voting app and didn't see the reason to change it.

src/components/Profile/Activity.js
fabriziovigevani: Ah so for the activity, we are removing duplicated stakes to the same proposal.

This is to prevent for example if i staked, withdrew, and staked again on a proposal, it doesn't show in your activity that you did all those 3 things and just show that you staked on the proposal in general since there's no real value in showing all those 3.

src/hooks/useAccountTotalStaked.js
fabriziovigevani: Yes! We are filtering by the actives ones only

src/utils/date-utils.js
fabriziovigevani: Ah it's the month in letters. https://github.com/iamkun/dayjs/blob/dev/docs/en/API-reference.md#list-of-all-available-formats

PR #117, opened 2020-10-14
Proposals paging fixed

famole:
After open a proposal and get back to the main screen, the list is set to the first page no matter in which page you were.
That was fixed!


fabriziovigevani (approved):
🚀

PR #119, opened 2020-10-15
Sort based on staked HNY

onbjerg:
During the proposal data refactor this part was missed: the proposals are no longer sorted correctly.

This fix is temporary since it does not sort on conviction (I could not figure out how to get the data in a simple manner), but on the amount of staked HNY.


lkngtn (approved):

0xGabi (approved):
Looks good to me, I think that filter for the amount of HNY staked is a good trade off. Probably the logic to filter the conviction is way to complicated

PR #123, opened 2020-10-16
Responsive

fabriziovigevani:
Left to do
- [x] Implement Filter slider on mobile
- [x] Profile 
- [x] Proposal detail


src/components/Feed/ProposalHeader.js
```diff
@@ -13,11 +21,13 @@ function ProposalHeader({ proposal, onSelectProposal }) {
...
     >
       <ProposalCreator proposal={proposal} />
-      <ContextMenu>
-        <ContextMenuItem onClick={onSelectProposal}>
-          <IconInfo /> Details
-        </ContextMenuItem>
-      </ContextMenu>
+      {layoutName !== 'small' && (
+        <ContextMenu>
+          <ContextMenuItem onClick={onSelectProposal}>
```
rperez89, line 26: wondering if we even need those menu items only for the details

src/components/Feed/ProposalsList.js
```diff
@@ -1,34 +1,72 @@
 import React, { useRef } from 'react'
-import { Button, GU } from '@1hive/1hive-ui'
+import {
+  BIG_RADIUS,
+  Button,
+  GU,
+  useTheme,
```
rperez89, line 6: Lets alphabetize this import

src/components/Filters/Filters.js
```diff
@@ -0,0 +1,137 @@
...
+  } = props
+
+  return (
+    <div
+      css={`
+        min-width: 270px;
+        height: fit-content;
+        margin-top: ${3 * GU}px;
+        margin-right: ${8 * GU}px;
+
```
rperez89, line 42: blank space

src/components/Layout.js
```diff
@@ -1,13 +1,16 @@
...
+function CustomLayout({ children }) {
+  const { width: vw } = useViewport()
   return (
     <Layout
       breakpoints={BREAKPOINTS}
-      paddingBottom={paddingBottom || 0}
+      parentWidth={vw}
+      paddingBottom={0}
       css={`
+        ${vw < BREAKPOINTS.large && `width: auto;`}
```
rperez89, line 13: i think here we might be adding `false` to the css because of the  `&&`  operator

rperez89 (commented):
LGTM!

src/components/Layout.js
fabriziovigevani: In this case it works the same as conditional rendering of react components so we don't need to worry about that.

src/components/Feed/ProposalHeader.js
fabriziovigevani: As discussed offline, the menu item is needed. We cannot do the proposal detail link be the whole card as we also have action buttons on it.

PR #125, opened 2020-10-16
Cache selected filters

fabriziovigevani:


0xGabi (approved):
Solid change! 

One nice to have is to also remember the scroll positon of the feed, but we can include on a later PR 👍

PR #133, opened 2020-10-20
Decision card

fabriziovigevani:


src/hooks/useExtendedVoteData.js
```diff
@@ -76,3 +69,25 @@ export default function useExtendedVoteData(vote) {
     startTimestamp,
   }
 }
+
+export function useCanUserVote(vote) {
```
fabriziovigevani, line 73: Had to move this into its own hook since we also use it for the Decision card actions

src/hooks/useExtendedVoteData.js
```diff
@@ -18,7 +18,7 @@ export default function useExtendedVoteData(vote) {
 
   const dandelionVotingContract = useContractReadOnly(
     dandelionVotingAddress,
-    dandelionVotingAbi.abi
+    dandelionVotingAbi
```
rperez89, line 21: does it work like this? i think that we have the abi key on the json

rperez89 (approved):
just one comment LGTM!

src/hooks/useExtendedVoteData.js
fabriziovigevani: Ah yes, i updated the json file to only include the abi part, the rest was not needed.

PR #135, opened 2020-10-20
Improvements

fabriziovigevani:
Does various improvements across the app.

More important ones: 

- Memoizes some components in order to avoid more re-renders.
- Memoizes proposal and config objects coming from the subscription (by stringifying the value).

### DIFF
--- a/src/components/DecisionDetail/VoteActions.js
+++ b/src/components/DecisionDetail/VoteActions.js
@@ -40,8 +40,6 @@ const VoteActions = React.memo(({ vote, onVoteYes, onVoteNo, onExecute }) => {
     startTimestamp,
   } = useExtendedVoteData(vote)
 
-  console.log('AFTER DATAAAAAAA', canUserVote, canExecute)
-
   const hasVoted = [VOTE_YEA, VOTE_NAY].includes(connectedAccountVote)
 
   useEffect(() => {
--- a/src/components/DecisionDetail/VoteActions.js
+++ b/src/components/DecisionDetail/VoteActions.js
@@ -46,14 +46,18 @@ const VoteActions = React.memo(({ vote, onVoteYes, onVoteNo, onExecute }) => {
     let cancelled = false
 
     const whenReady = async () => {
-      await Promise.all([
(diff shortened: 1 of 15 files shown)


src/components/DecisionDetail/VoteActions.js
```diff
@@ -40,22 +40,24 @@ const VoteActions = React.memo(({ vote, onVoteYes, onVoteNo, onExecute }) => {
...
     const whenReady = async () => {
-      await Promise.all([
-        canUserVotePromise,
-        canExecutePromise,
-        userBalancePromise,
-        userBalanceNowPromise,
-      ])
-      if (!cancelled) {
-        setReady(true)
+      try {
```
rperez89, line 49: ah, was this failing?  did you get to see the reason?

src/components/DecisionDetail/VoteCasted.js
```diff
@@ -1,13 +1,13 @@
...
+import { formatTokenAmount } from '../../lib/token-utils'
+import { getAccountCastStake } from '../../lib/vote-utils'
 import { VOTE_YEA } from '../../constants'
 
-function VoteCasted({ vote }) {
-  const { connectedAccountVote } = vote
-  const { userBalance } = useExtendedVoteData(vote)
+function VoteCasted({ account, accountVote, vote }) {
   const { config } = useAppState()
+  const accountStake = getAccountCastStake(vote, account)
```
rperez89, line 10: ah getting it from the hook was not working?

rperez89 (approved):
LGTM!

src/components/DecisionDetail/VoteActions.js
fabriziovigevani: Was having an error that could fix afterwards, but i also thought it would be a good idea to catch the errors

src/components/DecisionDetail/VoteCasted.js
fabriziovigevani: As discussed previously, the user's snapshot is not always correct as the users voting weight as they could have transferred some amount before voting

PR #144, opened 2020-10-21
Text search filter

### LINKED ISSUE #138: Search on v2?
There is no search on v2 but it is still a pretty nice feature to have, so if we can we should add one.

fabriziovigevani:
closes #138 

Adds functionality to filter proposals by name directly from the subgraph.

Uses new honeypot connector version with this filter functionality. See https://github.com/1Hive/honey-pot/pull/77/commits/2f70b3bb213d5615331ad730c83dbca44967032f

One **caveat** is that the subgraph text queries are case-sensitive.

Screen Shot 2020-10-21 at 1 20 09 PM


onbjerg (commented):
We can add fulltext search to the subgraph which will make it case insensitive: https://thegraph.com/docs/define-a-subgraph#defining-fulltext-search-fields

onbjerg (approved):
Seems like full text search does not work with filters, I will ask The Graph about that and until they fix it/if they do this is good

PR #145, opened 2020-10-21
Dandelion warning

rperez89:


src/components/Feed/ProposalFooter.js
```diff
@@ -137,8 +159,33 @@ function DecisionFooter({ proposal, onVoteOnDecision }) {
...
+      <Popover
+        visible={warningPopoverVisible}
+        opener={popoverOpener.current}
+        onClose={handleOnClosePopover}
+      >
+        <div
+          css={`
+            padding: ${3 * GU}px;
+            ${textStyle('body2')}
+            width:${compactMode ? 'auto' : 48 * GU}px;
```
fabriziovigevani, line 182: Let's add the border. Not sure if the color is part of the theme
```suggestion
            width:${compactMode ? 'auto' : 48 * GU}px;
            border: 1px solid #F5A623;
            border-radius: 16px;
```

src/components/Feed/ProposalFooter.js
```diff
@@ -137,8 +159,33 @@ function DecisionFooter({ proposal, onVoteOnDecision }) {
...
+          `}
+        >
+          Keep in mind that if you vote positive on Decisions, your entire HNY
+          balance will be locked until the end of the voting period.
```
fabriziovigevani, line 186: We should update this to the suggestion made by @willjgriff 

```
Voting in favour of a decision will prevent you from transferring your balance until it has been executed or x hours after the voting period ends. 

***For Devs:  x is `DandelionVoting.bufferBlocks() / 2 + execution delay` then converted from blocks to hours***
```

fabriziovigevani (commented):
Some comments! 

We should also add the warning in the [DecisionDetail view](https://www.figma.com/file/P06HTzSyb9hExJVpwaUPmA/Honey-Pot-2.0?node-id=1437%3A39)

PR #153, opened 2020-10-22
Small fixes

fabriziovigevani:
- Fixes Rankings component weird gap on mobile
- Wraps long words on the profile information.

### DIFF
--- a/src/components/Profile/MainProfile.js
+++ b/src/components/Profile/MainProfile.js
@@ -82,7 +82,7 @@ function MainProfile({ profile }) {
           >
             <div>
               {email && (
-                <div
+                <p
                   css={`
                     overflow: hidden;
                     white-space: nowrap;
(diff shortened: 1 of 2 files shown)


src/components/Profile/MainProfile.js
```diff
@@ -82,27 +82,25 @@ function MainProfile({ profile }) {
...
               )}
               {website && (
-                <div
+                <p
                   css={`
                     color: ${theme.contentSecondary};
-                    overflow: hidden;
-                    white-space: nowrap;
-                    text-overflow: ellipsis;
+                    word-break: break-word;
```
rperez89, line 99: shouldn't we keep the `...` on overflow?

rperez89 (commented):
LGTM just one small question

src/components/Profile/MainProfile.js
fabriziovigevani: In this case we are breaking the word, so we can be sure that it won't overflow.

e.g
**Before**
Screen Shot 2020-10-21 at 11 51 00 PM

**After**
Screen Shot 2020-10-21 at 11 52 20 PM

PR #155, opened 2020-10-22
Create Proposal->Founding->Warning box cutted off

famole:
When founding proposal was selected, the last box was cutted off as the image that I'm attaching. This is fixed on this PR.
Screen Shot 2020-10-22 at 12 08 56 AM

### DIFF
--- a/src/components/panels/AddProposalPanel.js
+++ b/src/components/panels/AddProposalPanel.js
@@ -189,7 +189,12 @@ const AddProposalPanel = React.memo(({ onSubmit }) => {
     errors.length > 0
 
   return (
-    <form onSubmit={handleFormSubmit}>
+    <form
+      onSubmit={handleFormSubmit}
+      css={`
+        margin-bottom: ${7 * GU}px;
+      `}
+    >
       <Field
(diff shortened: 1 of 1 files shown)


src/components/panels/AddProposalPanel.js
```diff
@@ -330,6 +330,7 @@ const AddProposalPanel = React.memo(({ onSubmit }) => {
           mode={neededThreshold ? 'info' : 'warning'}
           css={`
             margin-top: ${2 * GU}px;
+            margin-bottom: ${6 * GU}px;
```
fabriziovigevani, line 333: Let's set this margin to the `form` parent element, because at the bottom we have another warning component. Also lets set it to `${7 * GU}px` so the bottom spacing is even with the form upper spacing.

fabriziovigevani (commented):
Good catch @famole!

Left a comment

src/components/panels/AddProposalPanel.js
famole: Done!

fabriziovigevani (approved):
🙌

PR #161, opened 2020-10-22
V2 currency changer

anthonyoliai:
Currency changer that works for v2. I need some help on the switch case, what would be the best way to shorten it, so that it's less cluttered? Im still learning, I figured this was ugly code. It does work however.


anthonyoliai:
A preview of the currency changer can be found here.
https://gyazo.com/31ddb313af4e8ee680330fadd0126a17

onbjerg (commented):
[Here's a somewhat large diff](https://github.com/1Hive/honey-pot/compare/abe307b102e79a8337ecdd7074eac099e64f8e35..b4a0d6b6311b1e3c79b47b2865e87ddc2fc42331) with some proposed changes I have for simplifying the data model.

A short overview: instead of splitting symbols/names/rates, we just have a single `currency` object that we can pass around. This also simplifies the dropdown logic. A caveat on the commit I've pushed is that the default currency is no longer USD, but that is a minor thing we can easily fix.

If you want to apply the changes you can run:

```sh
git fetch
git cherry-pick b4a0d6b6311b1e3c79b47b2865e87ddc2fc42331
```

Let me know if you have questions! :sparkles:

anthonyoliai:
Changes are applied, with help of @onbjerg,  if one can have a review I would appreciate it :)

src/hooks/useOrgHooks.js
```diff
@@ -195,3 +195,64 @@ export function useTokenBalances(account, token, timer = 3000) {
...
+  const orderedCurrencies = orderCurrencies(currencies)
+  return orderedCurrencies
+}
+
+function orderCurrencies(currencies) {
+  const tempCurrency = currencies[0]
+  currencies[0] = currencies[8]
+  currencies[8] = tempCurrency
+  return currencies
+}
```
onbjerg, line 258: Just to make it a bit more "future-proof" in case they change the order of the currencies in their response :)

```suggestion
// Always puts USD at index 0
function orderCurrencies(currencies) {
  return currencies.sort(({ name }) => {
    if (name === 'USD') {
      return -1
    }
    return 0
  })
}
```

onbjerg (approved):
Other than the small nit looks good to me! We should wait for @fioreb to get back with design changes if any.

src/hooks/useOrgHooks.js
```diff
@@ -195,3 +195,64 @@ export function useTokenBalances(account, token, timer = 3000) {
...
+
+    fetchRates()
+
+    return () => {
+      cancelled = true
+      clearTimeout(retryTimer)
+    }
+  }, [])
+  const orderedCurrencies = orderCurrencies(currencies)
+  return orderedCurrencies
```
onbjerg, line 250:
```suggestion
  return orderCurrencies(currencies)
```

fioreb:
This changer is really cool! 😄

I thought that we can place it at the end of the box
https://www.figma.com/file/P06HTzSyb9hExJVpwaUPmA/Honey-Pot-2.0?node-id=1497%3A430
Let me know if it´s ok for you.
Thanks!

anthonyoliai:
Looks good to me!

> This changer is really cool! 😄
> 
> I thought that we can place it at the end of the box
> https://www.figma.com/file/P06HTzSyb9hExJVpwaUPmA/Honey-Pot-2.0?node-id=1497%3A430
> Let me know if it´s ok for you.
> Thanks!

src/hooks/useOrgHooks.js
```diff
@@ -195,3 +195,66 @@ export function useTokenBalances(account, token, timer = 3000) {
...
+  return orderCurrencies(currencies)
+}
+
+// Always puts USD at index 0
+function orderCurrencies(currencies) {
+  return currencies.sort(({ name }) => {
+    if (name === 'USD') {
+      return -1
+    }
+    return 0
```
onbjerg, line 258: Sorry, I think I lied to you. This has to be:

```suggestion
    return 1
```

and then it should work!

onbjerg (approved):
LGTM 5Head I'll let @rperez89 or @fabriziovigevani take a look and merge when they have time since they are the primary maintainers of Honey Pot

src/components/Metrics.js
```diff
@@ -15,7 +22,25 @@ const Metrics = React.memo(function Metrics({
 }) {
   const { layoutName } = useLayout()
   const compactMode = layoutName === 'small'
-  const { requestToken, stakeToken } = useAppState()
+  const { requestToken, stakeToken, currencies } = useAppState()
```
fabriziovigevani, line 25: It may make more sense to use the currency hook directly from this component as opposed to having it in the app state context since we will be using it only when the Metrics component is rendered. If in the future we see its useful to have as global state we could consider lifting the state up.

src/components/Metrics.js
```diff
@@ -15,7 +22,25 @@ const Metrics = React.memo(function Metrics({
...
+        symbol: '$',
+        rate: 1,
+      }
+    }
+
+    return currencies[currencyIndex]
+  }, [currencyIndex, currencies])
+  const handleCurrencyChange = useCallback(currencyIndex => {
+    setCurrencyIndex(currencyIndex)
+  }, [])
```
fabriziovigevani, line 40: This might not be needed. We can pass `setCurrencyIndex` directly to the dropdown `onChange` prop.

src/components/Metrics.js
```diff
@@ -15,7 +22,25 @@ const Metrics = React.memo(function Metrics({
 }) {
   const { layoutName } = useLayout()
   const compactMode = layoutName === 'small'
-  const { requestToken, stakeToken } = useAppState()
+  const { requestToken, stakeToken, currencies } = useAppState()
+  const [currencyIndex, setCurrencyIndex] = useState(0)
+  const currency = useMemo(() => {
+    if (!currencies.length) {
```
rperez89, line 28: maybe we can always have the usd currency right? i mean return always with the USD as default

src/components/Metrics.js
```diff
@@ -43,27 +68,42 @@ const Metrics = React.memo(function Metrics({
...
+            currency={currency}
+          />
+        </div>
+        <div>
+          <DropDown
+            header="Type"
+            placeholder="USD"
+            selected={currencyIndex}
+            onChange={handleCurrencyChange}
+            css={`
```
rperez89, line 103: we always try to place the css prop as the last one of the props, because sometimes is a big one

src/hooks/useOrgHooks.js
```diff
@@ -195,3 +195,66 @@ export function useTokenBalances(account, token, timer = 3000) {
 
   return balances
 }
+
+export function useCurrencies() {
```
fabriziovigevani, line 199: Related to the second comment, lets move this into its own file as it's not really an org hook. Consider also declaring the url and Symbol Map outside the hook function.

src/components/Metrics.js
```diff
@@ -96,11 +136,11 @@ function Metric({ label, value, color }) {
   )
 }
 
-function TokenBalance({ label, token, value }) {
+function TokenBalance({ label, token, value, currency }) {
   const theme = useTheme()
```
rperez89, line 141: here we could do `const {currencyRate, currencySymbol} = currency` and use those ones

src/components/Metrics.js
```diff
@@ -136,7 +176,8 @@ function TokenPrice() {
           color: ${theme.green};
         `}
       >
-        ${formatDecimals(price, 2)}
+        {currency.symbol}
```
rperez89, line 179: same suggestion here

fabriziovigevani (commented):
Awesome job @anthonyoliai!

Left some comments, let me know what do you think :)

src/providers/AppState.js
```diff
@@ -5,6 +5,7 @@ import {
   useOrgData,
   useTokenBalances,
   useVaultBalance,
+  useCurrencies,
```
rperez89, line 8: lets alphabetize the import

src/components/Metrics.js
```diff
@@ -1,5 +1,12 @@
-import React from 'react'
-import { Box, GU, textStyle, useLayout, useTheme } from '@1hive/1hive-ui'
+import React, { useState, useMemo, useCallback } from 'react'
+import {
+  Box,
+  GU,
+  textStyle,
+  useLayout,
+  useTheme,
+  DropDown,
```
rperez89, line 8: let's alphabetize this import :)

rperez89 (commented):
AWESOME work man!

anthonyoliai:
Hi! I'm so sorry if I accidentally clicked a wrong button. I'm not sure why it says I wanted to deploy. Regardless, I have worked on some of the suggestions. Putting setCurrencyIndex in the DropDown did not work, as there were too many rerenders so I kept it as is. Currently, the USD value is already set to default so I didn't tamper with that.

src/components/Metrics.js
```diff
@@ -1,9 +1,17 @@
...
+  GU,
+  textStyle,
+  useLayout,
+  useTheme,
+} from '@1hive/1hive-ui'
 
 import { useAppState } from '../providers/AppState'
 import { useUniswapHnyPrice } from '../hooks/useUniswapHNYPrice'
 import { formatDecimals, formatTokenAmount } from '../utils/token-utils'
+import { useCurrencies } from '../hooks/useCurrencies'
```
fabriziovigevani, line 14: One more small nitpick.

We generally order imports as well in the following order: 

 - External libraries
- Components
- Hooks
- Libraries
- Assets
```suggestion
import { useAppState } from '../providers/AppState'
import { useCurrencies } from '../hooks/useCurrencies'
import { useUniswapHnyPrice } from '../hooks/useUniswapHNYPrice'
import { formatDecimals, formatTokenAmount } from '../utils/token-utils'
```

fabriziovigevani (approved):
One small correction and LGTM! 🚀 

> . I'm not sure why it says I wanted to deploy.

Ah don't worry, every PR performs a preview deployment automatically now.

> Putting setCurrencyIndex in the DropDown did not work, as there were too many rerenders so I kept it as is.

Hmm that's odd, since `setCurrencyIndex` persist on every re-render, but yeah it's okay if we leave it as it is.


Awesome work!

fabriziovigevani (approved):

PR #162, opened 2020-10-22
Refactor use proposals

fabriziovigevani:
Refactors `useFilteredProposals` logic.

- Moves filtered decisions by status into the hook
- Exposes the total proposals fetched count that will be used to tell if we should fetch more proposals form the subgraph.


src/components/Feed/ProposalsList.js
```diff
@@ -17,6 +17,7 @@ function ProposalsList({
   activeFilters,
   proposals,
   proposalCount,
+  proposalsFetchedCount,
```
rperez89, line 20: isn't `proposalsFetchedCount` the same number than `filters.count.filter`?

rperez89:
We have a call where fabri explained me the changes LGTM!

PR #182, opened 2020-11-19
Withdraw stake

rperez89:
We noticed that given some concerns on the community about the abstain proposal stakes amounts that we were having some weird numbers on the active hny amounts, that is because we were not letting the user withdraw his stake from canceled or executed proposals.
At the beginning we were just relying on the ability of the smart contract to withdraw automatically from executed or canceled proposals the amount needed in case they want to support into another proposal but the reality is that if we don't give the chance to the user to withdraw it by themselves whenever they want is like we are forcing them to `Abstain` because those tokens staked on executed or cancelled proposals are taken into account as active because we are not automatically withdrawing in the smart contract for security reasons.

So a new component was added to the profile page where each user is going to be able to see in which proposals their have inactive stakes.

### DIFF
--- a/src/components/ProposalDetail/ProposalActions.js
+++ b/src/components/ProposalDetail/ProposalActions.js
@@ -10,6 +10,12 @@ import { addressesEqual } from '../../utils/web3-utils'
 import BigNumber from '../../lib/bigNumber'
 import { toDecimals, fromDecimals } from '../../utils/math-utils'
 
+import {
+  PROPOSAL_STATUS_ACTIVE_STRING,
+  PROPOSAL_STATUS_CANCELLED_STRING,
+  PROPOSAL_STATUS_EXECUTED_STRING,
+} from '../../constants'
+
 const MAX_INPUT_DECIMAL_BASE = 6
 
 function ProposalActions({
@@ -22,7 +28,7 @@ function ProposalActions({
   const { stakeToken, accountBalance } = useAppState()
   const { account: connectedAccount } = useWallet()
 
-  const { id, currentConviction, stakes, threshold } = proposal
+  const { id, currentConviction, stakes, threshold, status } = proposal
   const totalStaked = useAccountTotalStaked(connectedAccount)
 
   const myStake = useMemo(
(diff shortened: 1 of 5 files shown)


src/hooks/useProposals.js
```diff
@@ -139,6 +141,26 @@ export function useProposal(proposalId, appAddress) {
...
+  const { honeypot } = useAppState()
+
+  const supporter = useSupporterSubscription(honeypot, account)
+
+  if (!supporter || !supporter.stakes) {
+    return []
+  }
+  const inactiveStakes = supporter.stakes.filter(stake => {
+    return (
+      stake.proposal.type !== ProposalTypes.Decision &&
```
fabriziovigevani, line 155: Note that stakes are only part of Proposals and Suggestions so the type check is irrelevant here.

src/screens/Profile.js
```diff
@@ -35,6 +37,7 @@ function Profile() {
   const searchParams = useSearchParams()
   const selectedAccount = searchParams.get('account') || connectedAccount
   const accountStakes = useAccountStakes(selectedAccount)
+  const accountInactiveStakes = useInactiveProposalsWithStake()
```
fabriziovigevani, line 40: Would it make sense to use it inside the `InactiveProposalsStake` component directly?

src/hooks/useProposals.js
```diff
@@ -139,6 +141,26 @@ export function useProposal(proposalId, appAddress) {
...
+      stake.proposal.status !== PROPOSAL_STATUS_ACTIVE_STRING &&
```
fabriziovigevani, line 156: Should we check for Cancelled or Executed instead? As we do for checking the `withdraw` `mode` on `ProposalActions`?

src/components/ProposalDetail/ProposalActions.js
```diff
@@ -22,7 +28,7 @@ function ProposalActions({
   const { stakeToken, accountBalance } = useAppState()
   const { account: connectedAccount } = useWallet()
 
-  const { id, currentConviction, stakes, threshold } = proposal
+  const { id, currentConviction, stakes, threshold, status } = proposal
```
fabriziovigevani, line 31:
```suggestion
  const { id, currentConviction, stakes, status, threshold } = proposal
```

fabriziovigevani (commented):
Nicely done Ro!

Some minor comments

src/screens/Profile.js
rperez89: hmm not sure we do the same with the `stakingTokens` component, probably i would let like this for consistency

rperez89:
> Nicely done Ro!
> 
> Some minor comments
> 
> Also, wondering if we should put some spacing between each proposal item
> 
> Screen Shot 2020-11-20 at 4 18 04 PM

ah totally didn't realize about it because i only had one heheh

fabriziovigevani (approved):

PR #183, opened 2020-11-19
Profile wrap fix

anthonyoliai:
Fixes the wrapping problem on smaller resolutions on the profile page.


anthonyoliai:
It seems like the same problem persists on the main page which shows all the proposals. I will apply the same fix to the main page, and commit it tomorrow.

fabriziovigevani (approved):
LGTM! Good catch :)

PR #186, opened 2020-11-26
Disputable voting connector

### LINKED ISSUE #180: Disputable connector
We need to include all the changes needed to support the disputable subgrah schema and data

rperez89:
This pr removes all the attributes from dandelion voting and added the disputable voting ones.
There are a bunch of functions and that aragon had in their connector that i think we don't need but let's see, in case we need them we can add later
Like having the current settings and on cast functions https://github.com/aragon/connect/blob/master/packages/connect-disputable-voting/src/thegraph/connector.ts#L92-L110

this PR closes #180


connector/src/types.ts
```diff
@@ -36,12 +39,15 @@ export interface ConvictionConfigData {
 
 export interface VotingConfigData {
   id: string
-  token: TokenData
```
fabriziovigevani, line 39: Think i missed from the subgraph reviews that we remove the voting token. Where are we storing it now? Because i think it's going to be needed.

connector/src/models/CollateralRequirement.ts
```diff
@@ -0,0 +1,36 @@
+import { formatBn } from '../helpers'
+import { CollateralRequirementData, IHoneypotConnector } from '../types'
+
+export default class CollateralRequirement {
+  #connector: IHoneypotConnector
+
+  readonly id: string
+  readonly voteId: string
```
fabriziovigevani, line 8: Should this be `proposalId` ?

Think we also forgot to assign it on `constructor`

connector/src/types.ts
```diff
@@ -67,6 +73,7 @@ export interface StakeHistoryData {
 
 export interface CastData {
   id: string
+  proposalId: string
```
fabriziovigevani, line 76: We need to add this also in the `Cast` class

fabriziovigevani (commented):
Awesome work Ro!

Left some questions. Noticed also that it would not build for me due to some errors i suggested to fix. 

Was wondering also if we should update the `examples` folder so we can do some basic queries? Or are we planning to include tests?

Regarding the extra functions, i agree with keeping them out of this iteration and if we end up  needing them we can include them.

connector/src/models/Proposal.ts
```diff
@@ -49,28 +69,77 @@ export default class Proposal {
...
-    this.startBlock = data.startBlock
-    this.executionBlock = data.executionBlock
-    this.snapshotBlock = data.snapshotBlock
-    this.supportRequiredPct = data.supportRequiredPct
-    this.minAcceptQuorum = data.minAcceptQuorum
-    this.yea = data.yea
-    this.nay = data.nay
-    this.votingPower = data.votingPower
+    //voting data
+    this.settingId = data.settingId
```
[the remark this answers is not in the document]
rperez89: the disputable voting yes keeps an array of settings <https://github.com/aragon/aragon-apps/blob/88712912e87ea2ce99bffcee044e3811b4dae26c/apps/voting-disputable/contracts/DisputableVoting.sol#L132> and they emit an newSettings event <https://github.com/aragon/aragon-apps/blob/88712912e87ea2ce99bffcee044e3811b4dae26c/apps/voting-disputable/contracts/DisputableVoting.sol#L138>

connector/src/thegraph/parsers/supporter.ts
```diff
@@ -4,7 +4,7 @@ import { CastData, IHoneypotConnector, StakeData, StakeHistoryData, SupporterDat
 
 export function parseSupporter(
   result: QueryResult,
-  connector: IHoneypotConnector
```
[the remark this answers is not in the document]
rperez89: yes the linter was marking as read pluse we are not using it in the others parsers

rperez89:
> Awesome work Ro!
> 
> Left some questions. Noticed also that it would not build for me due to some errors i suggested to fix.
> 
> Was wondering also if we should update the `examples` folder so we can do some basic queries? Or are we planning to include tests?
> 
> Regarding the extra functions, i agree with keeping them out of this iteration and if we end up needing them we can include them.

ah yes sorry i forgot to mention that didn't fully tested yet, wanted to know first i i was on the right path

fabriziovigevani (approved):
LGTM!

PR #206, opened 2021-02-03
Support proposal multimodal

famole:
First draft of the support proposal multimodal


src/screens/ProposalDetail.js
```diff
@@ -360,17 +364,16 @@ function ProposalDetail({
...
+        visible={miltiModalVisible}
+        onClose={() => setMiltiModalVisible(false)}
       >
-        <SupportProposalPanel
+        <SupportProposalScreens
           id={id}
-          onDone={panelState.requestClose}
-          onStakeToProposal={convictionActions.stakeToProposal}
+          onDone={() => setMiltiModalVisible(false)}
+          onStakeToProposal={actions.stakeToProposal}
```
famole, line 374: @rperez89 heres is where i need your help, at the moment of calling onStakeToProposal fails because onStakeToProposal is not a function.

rperez89: STEP 1
Another thing that you need to do is change the way that the stakeToProposal makes the contract call since now we have a modal transition that keeps loading while the trx is being mined as opposed to just propmpt metamask to sign as we were doing before, see here: https://github.com/1Hive/honey-pot/blob/3185f4b1c3abfb07cf89b3da4fc1ba64dbb709b6/src/hooks/useActions.js#L82-L92 there you can see that the `sendIntent` function ask ethers to get the signer what is the action that prompts metamask.
Now the modal is the one in charge of executing transactions https://github.com/1Hive/honey-pot/blob/disputable-version/src/components/ModalFlows/ModalFlowBase.js#L30-L74.
So we need to adapt the stakeToProposal action to do not call metamask and just generate an intent, an intent is kind of a pre generation of the data that needs to be send to the signer with the `from, to, data` that is going to be sent to the smart contract and you can see an example of how to do this here https://github.com/1Hive/honey-pot/blob/disputable-version/src/hooks/useActions.js#L172-L190

src/components/ModalFlows/SupportProposal/SupportProposalScreens.js
```diff
@@ -0,0 +1,52 @@
...
+import SupportProposal from './SupportProposal'
+import useActions from '../../../hooks/useActions'
+
+function SupportProposalScreens({ versionId, id, onDone, stakeProposal }) {
+  const actions = useActions()
+  const [transactions, setTransactions] = useState([])
+
+  const getTransactions = useCallback(
+    async onComplete => {
+      await actions.agreementActions.signAgreement({ versionId }, intent => {
```
rperez89, line 13: Step: 2 after changing the action you also need to change here to call the new action

src/components/ModalFlows/SupportProposal/SupportProposal.js
```diff
@@ -0,0 +1,186 @@
...
+    stakeToken,
+    accountBalance
+  )
+
+  // Form submit handler
+  const handleSubmit = useCallback(
+    event => {
+      event.preventDefault()
+
+      onStakeToProposal(id, amount.toString(10))
```
rperez89, line 41: STEP 3:
so here you need to instead of call the action you need to call the `getTransaction` that you are going to pass from the supportPRoposalScreens and move to the next modal screen, see this example https://github.com/1Hive/honey-pot/blob/3185f4b1c3abfb07cf89b3da4fc1ba64dbb709b6/src/components/ModalFlows/SignAgreementScreens/SignOverview.js#L16-L23

rperez89 (commented):
left a step by step of how to handle the transactions

src/hooks/useActions.js
```diff
@@ -80,15 +80,20 @@ export default function useActions(onDone) {
   )
 
   const stakeToProposal = useCallback(
-    (proposalId, amount) => {
-      sendIntent(convictionVotingApp, 'stakeToProposal', [proposalId, amount], {
-        ethers,
-        from: account,
-      })
+    async ({ proposalId, amount }, onDone = noop) => {
+      const intent = await agreementApp.intent(
```
rperez89, line 84: here is where you are having the issue, is not the agreement app we stake with the `convictionVotingApp`

src/components/ModalFlows/SupportProposal/SupportProposalScreens.js
```diff
@@ -1,39 +1,29 @@
 import React, { useMemo, useState, useCallback } from 'react'
-import PropTypes from 'prop-types'
 import ModalFlowBase from '../ModalFlowBase'
 import SupportProposal from './SupportProposal'
-import useActions from '../../../hooks/useActions'
 
-function SupportProposalScreens({ versionId, id, onDone, stakeProposal }) {
-  const actions = useActions()
+function SupportProposalScreens({ onStakeToProposal, id }) {
```
rperez89, line 5: here you can use the `useAction` hook and get the convictionVotingActions as we were doing in the line right bellow

src/components/ModalFlows/SupportProposal/SupportProposalScreens.js
```diff
@@ -1,39 +1,29 @@
...
         graphicHeader: false,
-        content: (
-          <SupportProposal
-            id={id}
-            getTransactions={getTransactions}
-            onDone={onDone}
-            onStakeProposal={stakeProposal}
-          />
-        ),
+        content: <SupportProposal id={id} getTransactions={getTransactions} />,
```
rperez89, line 23: Not sure if you need to pass the id to this component since you already have it in this component at the moment of calling the `stakeToProposal` here https://github.com/1Hive/honey-pot/pull/206/files/1dd37543ec0308d33c78da22fe5251c2d86386c2..faa8b3902b58242e2e79ceb730d6ca4b4593e202#diff-5ecbb913608e7df708a2dc6c7f21e7e034675b15e4a9ba11aa8f277cbcce968bR10 , is the id that you have in the propoerties, so here you are passing it to the `<SupportProposal>` component to pass it back to the parent here https://github.com/1Hive/honey-pot/pull/206/files/1dd37543ec0308d33c78da22fe5251c2d86386c2..faa8b3902b58242e2e79ceb730d6ca4b4593e202#diff-5daa6f474d614ca00802abd12c3e990e03506232d87e08dd10a30f0964c2c7dfR48

src/screens/ProposalDetail.js
```diff
@@ -379,10 +379,28 @@ function ProposalDetail({
...
+              challengeAction: actions.challengeAction,
+              getAllowance: actions.getAllowance,
+              approveChallengeTokenAmount: actions.approveChallengeTokenAmount,
+            }}
+            proposal={proposal}
+          />
+        )}
+        {modalMode === 'settle' && (
+          <SettleProposalScreens proposal={proposal} />
+        )}
```
rperez89: and i think that you need to add the `<SupportProposal>` screens here to call the modal with mode `stake` or something like that

rperez89 (changes requested):

PR #305, opened 2021-03-16
QA suggestions

### LINKED ISSUE #290: Suggestions
On the staking screen when depositing collateral, the warning should read: "You will be able to withdraw your collateral balance at any time, except for any that is locked on a scheduled proposal."
Screenshot 2021-03-12 at 15 11 56

This proposal is cancelled yet I can still stake on it. #17 for reference. If I try to stake it fails. Presumably we don't need to let the user explicitly unstake because anything in cancelled proposals should be take into account on the UI? In which case can we remove the support button in this situation?
Screenshot 2021-03-12 at 15 25 44

No table entry added to the stake screen for the previously cancelled proposal. The available Honey has updated but there's no indication as to why. Turns out I need to refresh and reconnect my wallet before it updates.
Screenshot 2021-03-12 at 15 35 00

The warning on the right for a suggestion should read: "The suggestion will exist until it is challenged or removed."
Screenshot 2021-03-12 at 15 50 46

Can we remove the ` at the start of the error message at the bottom.
Screenshot 2021-03-11 at 13 53 41

"remaining" should be "remainder".
Screenshot 2021-03-11 at 13 58 31

When challenging an action this needs to be laid out better. 
The action collateral message should remove the "Your wallet balance is 18 HNY."
The dispute fees message should remove the "This balance will be returned to your account if the submitter accepts your settlement offer or if you win the dispute raised to Celeste."
The dispute fees warning should say "lock" instead of "pay".
There should be an info box at the bottom saying "The action collateral and dispute fees will be returned to your account if the submitter accepts your settlement offer or if you win the dispute raised to Celeste. Your wallet balance is 18 HNY."
Screenshot 2021-03-12 at 16 10 56

Should read "This proposal has been challenged but you can continue to support it to increase conviction. When the challenge is resolved, if allowed, the proposal will be resumed otherwise this proposal will be cancelled."
Screenshot 2021-03-12 at 16 27 15

This should read "If you specify the proposal amount in DAI it will be converted to HNY at the time of execution."
Screenshot 2021-03-12 at 16 41 37

The links to the proposal from the staking manager don't work for me. They just go to home.

The modals disappear regularly. I've noticed if I click in the modal and remove the click after dragging the mouse outside of it, it will disappear. Can this be prevented? Of course clicking outside should still close it.

Also the modals don't disappear after a transaction has been processed, you have to click them away. I think they should close automatically.

The links to celeste are broken because it uses "questions" in the URL instead of "disputes".

I've noticed when depositing in the staking manager my transaction sometimes fails because the gas limit is set too low. Can it default to a higher value? Tx: https://rinkeby.etherscan.io/tx/0xf94d4f4b10e38eb5d42a9e18cd9a9688072e7a6f9f8b91c1855f6e321686d881

rperez89:
This closes #290 

Still need to address the modal one that will suggested but need to give a thought to it

### DIFF
--- a/src/components/ProposalDetail/ProposalActions.js
+++ b/src/components/ProposalDetail/ProposalActions.js
@@ -100,11 +100,10 @@ function ProposalActions({
     onRequestSupportProposal,
   ])
 
-  if (!connectedAccount) {
-    return <AccountNotConnected />
-  }
-
   if (mode) {
+    if (!connectedAccount) {
+      return <AccountNotConnected />
+    }
     return (
       <div>
         <Button
--- a/src/components/ProposalDetail/ProposalActions.js
+++ b/src/components/ProposalDetail/ProposalActions.js
@@ -42,6 +42,9 @@ function ProposalActions({
     if (didIStake) {
       return 'update'
     }
+    if (hasEnded) {
+      return null
+    }
     return 'support'
   }, [currentConviction, didIStake, hasEnded, threshold])
 
@@ -77,12 +80,15 @@ function ProposalActions({
         mode: 'strong',
       }
     }
-    return {
-      text: 'Support this proposal',
-      action: onRequestSupportProposal,
-      mode: 'strong',
-      disabled: !accountBalance.gt(0),
+    if (mode === 'support') {
+      return {
+        text: 'Support this proposal',
+        action: onRequestSupportProposal,
+        mode: 'strong',
+        disabled: !accountBalance.gt(0),
+      }
     }
+    return null
   }, [
     accountBalance,
     handleExecute,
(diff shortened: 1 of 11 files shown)


src/screens/Home.js
```diff
@@ -145,6 +145,7 @@ const Home = React.memo(function Home() {
           <MultiModal
             visible={createProposalModalVisible}
             onClose={() => setCreateProposalModalVisible(false)}
+            onClosed={() => setCreateProposalModalVisible(false)}
```
fabriziovigevani, line 148: Is this necessary? Seems that is doing the same as `onClose` which will be fired first?

src/components/ProposalDetail/ProposalActions.js
```diff
@@ -77,12 +80,15 @@ function ProposalActions({
...
-      disabled: !accountBalance.gt(0),
+    if (mode === 'support') {
+      return {
+        text: 'Support this proposal',
+        action: onRequestSupportProposal,
+        mode: 'strong',
+        disabled: !accountBalance.gt(0),
+      }
     }
+    return null
```
fabriziovigevani, line 91: Should we place it at the top to avoid all the if statements?

```suggestion
   if(!mode) {
     return null
   }
```

fabriziovigevani (commented):
Awesome Ro!

Some comments


> Still need to address the modal one that will suggested but need to give a thought to it
If we end up closing the issue #290, let's create a new one with this pending task

fabriziovigevani (approved):
Awesome!

PR #311, opened 2021-03-17
new deployment data

rperez89:


src/hooks/useOrgHooks.js
```diff
@@ -111,11 +111,11 @@ export function useOrgData() {
...
 
-export function useVaultBalance(installedApps, token, timeout = 1000) {
-  const vaultAddress = getAppAddressByName(installedApps, 'vault')
-  const vaultContract = useContractReadOnly(vaultAddress, vaultAbi)
+export function useAgentBalance(installedApps, token, timeout = 1000) {
+  const agentAddress = getAppAddressByName(installedApps, 'agent')
+  const agentContract = useContractReadOnly(agentAddress, agentAbi)
 
-  const [vaultBalance, setVaultBalance] = useState(new BigNumber(-1))
+  const [agentBalance, setAgentBalance] = useState(new BigNumber(-1))
```
fabriziovigevani, line 118: There's no need to change to agent, what we could do instead is search for the vault address by `vault` as the appName, if it's not present, then get it with `agent`

We don't need to import the agent abi either, the `vault-balance` file we are using will do just fine since we are using the `balance` function only.

```
  let vaultAddress = getAppAddressByName(installedApps, 'vault')
if (!vaultAddress) {
   vaultAddress = getAppAddressByName(installedApps, 'agent')
}
  const agentContract = useContractReadOnly(vaultAddress, vaultAddress)
```

Issue #329, opened 2021-03-19
Some suggestions (mainly copy chnages)


willjgriff:
Get Honey link has Honey as input currency. It should be the output currency. Use this link for the "Get Honey" link: 

`https://app.honeyswap.org/#/swap?outputCurrency=0x71850b7e9ee3f13ab46d67167341e4bdc905eef9`

I think these buttons should be the other way round @fioreb ? Deposit is always the first action someone does so should it not be at the top?
<img width="345" alt="Screenshot 2021-03-19 at 10 59 39" src="https://user-images.githubusercontent.com/9317743/111770349-35957980-88a2-11eb-911f-2cf181b22e89.png">

I think we should replace the term "action" with "proposal" in every string visible in this modal. Also update "staking manager" to "collateral manager". 

And can we align the money image with the "Action Deposit" label so its centred horizontally?

<img width="685" alt="Screenshot 2021-03-19 at 11 06 52" src="https://user-images.githubusercontent.com/9317743/111771142-38dd3500-88a3-11eb-8665-951699c2fb24.png">

Similarly can we update the title of this modal to "Submit proposal requirements"

Also the Covenant Signature and Version string should be "You must sign the community covenant in order to create a proposal. The Covenant was last updated on 17/03/2021." Don't miss the change from "at 17/03/2021" to "on 17/03/2021." btw.

The Action Collateral string should be "You must lock 0.1 HNY as collateral to create this proposal. Your current collateral manager balance is 1.09 HNYT."
<img width="687" alt="Screenshot 2021-03-19 at 11 27 32" src="https://user-images.githubusercontent.com/9317743/111773455-1bf63100-88a6-11eb-9771-bda7e4e37e3b.png">

Change "signaling proposal" to "suggestion".
<img width="609" alt="Screenshot 2021-03-19 at 11 34 59" src="https://user-images.githubusercontent.com/9317743/111774269-2664fa80-88a7-11eb-8e3c-90d7af78a6dd.png">

Update to "Proposals cannot be challenged more than once."
<img width="306" alt="Screenshot 2021-03-19 at 11 58 30" src="https://user-images.githubusercontent.com/9317743/111776879-6f6a7e00-88aa-11eb-9350-693a20394b11.png">

This should read "Action Collateral is the amount of tokens locked every time an action is created. This amount will be locked in the collateral manager provided access has been granted."
[...]

PR #333, opened 2021-03-19
Suggestions

fabriziovigevani:
Partially addresses #329 

Still need to define whether we want to update `action` to `proposal` in Proposal creation modals or not as there hasn't been a clear consensus


rperez89 (commented):
LGTM, wondering if we also can add here the suggestion made by will about renaming this https://discord.com/channels/698287700834517064/747839698814107699/822732622664826900

rperez89 (approved):
LGTM!

PR #335, opened 2021-03-19
Fixed issue on mobile resolution

famole:
Screen Shot 2021-03-19 at 6 28 31 PM
Screen Shot 2021-03-19 at 6 28 13 PM
Screen Shot 2021-03-19 at 6 24 24 PM

### DIFF
--- a/src/components/ModalFlows/CreateProposalScreens/ActionFees.js
+++ b/src/components/ModalFlows/CreateProposalScreens/ActionFees.js
@@ -51,98 +51,56 @@ function ActionFeesModal({ agreement, onCreateTransaction }) {
         </span>
       </Field>
 
-      {compactMode && (
-        <div
-          css={`
-            margin-top: ${GU * 2}px;
-            weight: 100%;
-          `}
-        >
-          <img
-            css={`
-              height: ${GU * 4}px;
-              width: ${GU * 4}px;
-            `}
-            src={iconFees}
-            alt=""
-          />
-        </div>
-      )}
-
       <div
         css={`
           display: flex;
           width: 100%;
           height: ${GU * 4}px;
           margin-top: ${!compactMode ? GU * 3 : 0}px;
+          align-items: center;
         `}
       >
-        {!compactMode && (
-          <div
-            css={`
-              flex: 0.25;
-            `}
-          >
-            <img
-              css={`
-                height: ${GU * 4}px;
-                width: ${GU * 4}px;
-              `}
-              src={iconFees}
-              alt=""
-            />
-          </div>
-        )}
+        <img
+          css={`
+            height: ${GU * 4}px;
+            width: ${GU * 4}px;
+          `}
+          src={iconFees}
+          alt=""
+        />
 
-        <div
+        <h3
           css={`
-            flex: ${compactMode ? 1 : 2};
-            padding: 0 0 0 ${!compactMode ? 1.25 * GU : 0}px;
-            align-self: center;
+            font-weight: 600;
+            margin-left: 3%;
           `}
         >
-          <h3
-            css={`
-              font-weight: 600;
-            `}
-          >
-            Action Deposit
-          </h3>
-        </div>
-        <div
+          Action Deposit
+        </h3>
+
+        <span
           css={`
-            flex: 0.5;
-            align-self: center;
-            ${compactMode ? textStyle('body4') : textStyle('body3')};
+            text-align: 'left';
+            font-weight: 600;
+            position: absolute;
+            right: ${compactMode ? GU * 13.1 : GU * 19}px;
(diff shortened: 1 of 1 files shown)


src/components/ModalFlows/CreateProposalScreens/ActionFees.js
```diff
@@ -51,98 +51,56 @@ function ActionFeesModal({ agreement, onCreateTransaction }) {
...
-              alt=""
-            />
-          </div>
-        )}
+        <img
+          css={`
+            height: ${GU * 4}px;
+            width: ${GU * 4}px;
+          `}
+          src={iconFees}
```
fabriziovigevani, line 68: Lets use the `img` properties 

```suggestion
          css={`
            height: ${GU * 4}px;
            width: ${GU * 4}px;
          `}
          src={iconFees}
          height={4 * GU}
          width={4 * GU}
```

src/components/ModalFlows/CreateProposalScreens/ActionFees.js
```diff
@@ -51,98 +51,56 @@ function ActionFeesModal({ agreement, onCreateTransaction }) {
...
+        />
 
-        <div
+        <h3
           css={`
-            flex: ${compactMode ? 1 : 2};
-            padding: 0 0 0 ${!compactMode ? 1.25 * GU : 0}px;
-            align-self: center;
+            font-weight: 600;
+            margin-left: 3%;
```
fabriziovigevani, line 75: Lets use px here since i think making it 8px margin should be okay 
```suggestion
            margin-left: ${GU}px;
```

src/components/ModalFlows/CreateProposalScreens/ActionFees.js
```diff
@@ -51,98 +51,56 @@ function ActionFeesModal({ agreement, onCreateTransaction }) {
...
-        <div
+          Action Deposit
+        </h3>
+
+        <span
           css={`
-            flex: 0.5;
-            align-self: center;
-            ${compactMode ? textStyle('body4') : textStyle('body3')};
+            text-align: 'left';
```
fabriziovigevani, line 83:
```suggestion
            text-align: left;
```

src/components/ModalFlows/CreateProposalScreens/ActionFees.js
```diff
@@ -51,98 +51,56 @@ function ActionFeesModal({ agreement, onCreateTransaction }) {
...
+            font-weight: 600;
+            position: absolute;
```
fabriziovigevani, line 85: Ah lets not use abolsute if we don't need to.

If we want to separate the dollarAmount and token amount from the icon and text we can do so by encapsulating them in a div and in the parent div use `justify-content: space-between`.

A code snippet of how it would look like would be

```
<div
  css={`
    width: 100%;
    height: ${GU * 4}px;
    margin-top: ${!compactMode ? GU * 3 : 0}px;
    display: flex;
    align-items: center;
    justify-content: space-between;
  `}
>
  <div
    css={`
      display: flex;
      align-items: center;
    `}
  >
    <img
      css={`
        height: ${GU * 4}px;
        width: ${GU * 4}px;
      `}
      src={iconFees}
      alt=""
    />

    <h3
      css={`
        font-weight: 600;
        margin-left: ${1 * GU}px;
      `}
    >
      Action Deposit
    </h3>
  </div>
  <div>
    <span
      css={`
        text-align: left;
        font-weight: 600;
        ${compactMode ? textStyle('body3') : textStyle('body2')};
        margin-right: ${(compactMode ? 1.5 : 3) * GU}px;
      `}
    >
      ${dollarAmount}
    </span>

    <span
      css={`
        font-weight: 600;
        ${compactMode ? textStyle('body3') : textStyle('body2')};
      `}
    >
      {formatedAmount} {token.symbol}
    </span>
  </div>
</div>

```

fabriziovigevani (commented):
Awesome Fabi!

Left some comments including how we could restructure a bit the row so we don't use absolute positions. Let me know what do you think.

Also, to keep consistency, could we update everywhere we use `${GU * X}px` for `${X * GU}px` since that's how we do it in the rest of the code base.

src/components/ModalFlows/CreateProposalScreens/ActionFees.js
famole: Hey @fabriziovigevani thanks for the clarification on this, i've just applied the changes.
Also the GU pattern was my bad, i'll take that in mind for the future :)

fabriziovigevani: No worries! Thanks for updating them!

fabriziovigevani (approved):
LGTM!

PR #339, opened 2021-03-23
Remove Sign Covenant button if already signed

fabriziovigevani:
Also updates button loading state when modal closed


rperez89 (approved):
LGTM!

PR #343, opened 2021-03-24
Create proposal redirection

fabriziovigevani:
- Renders `Create proposal` button after txs are completed
- Updates `stake` route for `collateral`
- Handles Proposal creation modal visibility when route includes `/create`
- Does some fixes suggested at https://github.com/1Hive/honey-pot/pull/342

### DIFF
--- a/src/components/ModalFlows/ModalFlowBase.js
+++ b/src/components/ModalFlows/ModalFlowBase.js
@@ -98,10 +98,6 @@ function ModalFlowBase({
           <Stepper
             steps={transactionSteps}
             onCompleteActions={onCompleteActions}
-            css={`
-              margin-top: ${3.25 * GU}px;
-              margin-bottom: ${(onCompleteActions ? 0 : 5.5) * GU}px;
-            `}
           />
         ),
       })
--- a/src/components/ModalFlows/ModalFlowBase.js
+++ b/src/components/ModalFlows/ModalFlowBase.js
@@ -22,6 +22,7 @@ function ModalFlowBase({
   screens,
   transactions,
   transactionTitle,
+  onCompleteActions,
 }) {
   const { ethers } = useWallet()
   const signer = useMemo(() => ethers.getSigner(), [ethers])
@@ -96,9 +97,10 @@ function ModalFlowBase({
         content: (
           <Stepper
             steps={transactionSteps}
+            onCompleteActions={onCompleteActions}
             css={`
               margin-top: ${3.25 * GU}px;
-              margin-bottom: ${5.5 * GU}px;
+              margin-bottom: ${(onCompleteActions ? 2 : 5.5) * GU}px;
             `}
           />
         ),
@@ -113,6 +115,7 @@ function ModalFlowBase({
     transactionTitle,
     loading,
     frontLoad,
+    onCompleteActions,
   ])
 
   return <MultiModalScreens screens={extendedScreens} />
--- a/src/components/ModalFlows/ModalFlowBase.js
+++ b/src/components/ModalFlows/ModalFlowBase.js
@@ -100,7 +100,7 @@ function ModalFlowBase({
             onCompleteActions={onCompleteActions}
             css={`
               margin-top: ${3.25 * GU}px;
-              margin-bottom: ${(onCompleteActions ? 2 : 5.5) * GU}px;
+              margin-bottom: ${(onCompleteActions ? 0 : 5.5) * GU}px;
             `}
           />
         ),
--- a/src/components/ModalFlows/StakeScreens/StakeScreens.js
+++ b/src/components/ModalFlows/StakeScreens/StakeScreens.js
@@ -1,4 +1,6 @@
 import React, { useCallback, useMemo, useRef, useState } from 'react'
+import { useHistory } from 'react-router'
+import { Button } from '@1hive/1hive-ui'
 import ModalFlowBase from '../ModalFlowBase'
 import StakeAndWithdraw from './StakeAndWithdraw'
 import { useAppState } from '../../../providers/AppState'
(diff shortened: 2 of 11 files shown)


src/components/ModalFlows/ModalFlowBase.js
```diff
@@ -96,10 +97,7 @@ function ModalFlowBase({
         content: (
           <Stepper
             steps={transactionSteps}
-            css={`
```
rperez89, line 99: are we changing this for some reason?

src/components/ModalFlows/StakeScreens/StakeScreens.js
```diff
@@ -27,13 +46,27 @@ function StakeScreens({ mode, stakeManagement, stakeActions }) {
...
-          onComplete()
+          temporatyTrx.current = temporatyTrx.current.concat(intent)
+
+          if (stakingAllowance?.gt(0)) {
+            setTransactions(temporatyTrx.current)
+            onComplete()
+          }
         })
+
+        if (stakingAllowance?.eq(0)) {
```
rperez89, line 59: giving a second thought to this and looking at it i think that is a bit redundat to have this other if here

src/components/ModalFlows/StakeScreens/StakeScreens.js
```diff
@@ -10,9 +12,26 @@ const ZERO_BN = new BigNumber(toDecimals('0', 18))
...
+  const history = useHistory()
+  const { allowance: stakingAllowance } = stakeManagement?.staking || {}
 
   const temporatyTrx = useRef([])
 
+  const handleCreateProposal = useCallback(() => {
+    history.push('/home/create')
+  }, [history])
+
+  const renderOnCompleteActions = useCallback(() => {
```
rperez89, line 24: wondering if we should have the button on the modal instead of creating the component here and passsing down to it, thinking about passing only the button props. Or maybe you think on this for the possibility for the modal to show something else than buttons?

src/components/Stepper/Stepper.js
```diff
@@ -224,6 +249,7 @@ function Stepper({ steps, onComplete, ...props }) {
           {layout === 'expanded' && renderSteps()}
         </ul>
       </div>
+      {renderNextActions()}
```
rperez89, line 252: if the answer for the previous question is that you thought about it to render anything i think that might make sense to have the if inside the html as we do always instead of having the function call

src/hooks/useStaking.js
```diff
@@ -334,17 +334,40 @@ export function useStaking() {
...
+            data: allowManagerData,
+            from: account,
+            to: stakeManagement.stakingInstance,
+            description: 'Allow 1Hive Protocol',
+          },
+        ]
+        if (mounted()) {
+          onDone(intent)
+        }
+      } else {
```
rperez89, line 361: are we starting to avoid early returns now for some reason?

rperez89 (commented):
awesome fabri, just some discussions

src/components/ModalFlows/ModalFlowBase.js
fabriziovigevani: Yes, since we might render more components, the margins should be adjusted accordingly.

https://github.com/1Hive/honey-pot/pull/343/files#diff-4b485407c66360ee14514f9baad0c3f18a50297893c63447d20a044007854399R176

src/components/ModalFlows/StakeScreens/StakeScreens.js
fabriziovigevani: Yeah i think this approach is a general one where tyou can render anything you'd like.

Also the problem with rendering on the modal is that the modal doesn't control the transactions statuses so it doesn't know when it should render, unless we lift the transaction steps status logic to the modal which don't think makes sense really.

src/components/Stepper/Stepper.js
fabriziovigevani: I preferred to call the function and return null if conditions are not fulfilled so we don't have too much logic at the render level

src/hooks/useStaking.js
fabriziovigevani: No, but as i commented on your PR i think it's more clear for a new dev jumping in the codebase, as an early return could be overlooked in this case.

Also when it comes to performing logic conditionally i think we should always handle it in this way and only do early returns when we are returning data: e.g as we do in useMemos

src/components/ModalFlows/StakeScreens/StakeScreens.js
fabriziovigevani: Maybe refactoring it this way might be more clear

```
   await stakeActions.stake({ amount }, intent => {
        temporatyTrx.current = temporatyTrx.current.concat(intent)
      })
    
      if (stakingAllowance?.eq(0)) {
        await stakeActions.allowManager(intent => {
          temporatyTrx.current = temporatyTrx.current.concat(intent)
        })
      }
    
      setTransactions(temporatyTrx.current)
      onComplete()
    
      return
```

src/components/Stepper/Stepper.js
rperez89: i got your point for when we have a really complex logic, but in this case does not apply, we barely have html rendered on functions in our codbase and the good thing about having all in the html is that you don't need to take out your view to the logic part of the component to follow the html tree. so in this case i would stick as we usually do

src/hooks/useStaking.js
rperez89: got it

src/components/ModalFlows/StakeScreens/StakeScreens.js
rperez89: make sense!

src/components/Stepper/Stepper.js
fabriziovigevani: Ok that's fair!

Issue #336, opened 2021-03-22
Changes for the Create Proposal flow


fioreb:
There is some stuff that we need to change for the cases when the user is going to create a proposal but didn't have the Covenant Signed / didn't deposit collateral / didn't allow the lock manager.

Here is the figma with the complete flow:
https://www.figma.com/file/CrXSadZeZk8M4LO1jNLWrZ/HNYPot-Flow-check?node-id=0%3A1

Changes that we need:

### **COVENANT PAGE CHANGES:**
- We should remove the "Sign Covenant button" if the user already signed the covenant instead of showing it disabled.

### **SIGN COVENANT MODAL CHANGES:**
- Add the "Create proposal" button once the transaction is completed

### **COLLATERAL MANAGER PAGE CHANGES:**
- Change "Collateral Management" to "Collateral Manager"

- Invert the order of the buttons in the Collateral Manager page, Deposit should be first.

- Change "Allow Lock Manager" of the toggle text to "Allow 1Hive Protocol"

### **SUBMIT ACTION REQUIREMENTS MODAL CHANGES:**

- Stake collateral button should change to "Deposit collateral" (I think this change is already done, but just in case)

### **DEPOSIT HNY AS COLLATERAL MODAL CHANGES:**
- We should add a third step to the modal for the user to allow the 1Hive protocol

- We need to make some color adjustments to differentiate more the states of "Transaction is being minded..." and "Transaction successfully mined", [here](https://www.figma.com/file/CrXSadZeZk8M4LO1jNLWrZ/HNYPot-Flow-check?node-id=8%3A4) you can see the differentiation. Not only the icon changes, the color of the arrow too, they should be yellow when the transaction is being minded. The text should change the color too

- The "Transaction being minded" changes to "Transaction is being minded..."

- The "Transaction mined" text changes to "Transaction successfully mined"

- The Waiting for signature step indicator, should change the grey to our grey, now it has the Aragon's one. [here is the design](https://www.figma.com/file/CrXSadZeZk8M4LO1jNLWrZ/HNYPot-Flow-check?node-id=6%3A3613)

- Add the "Create proposal" button once the last transaction is completed

________

I added a file with the assets that we need to use for the changes, for the Transaction successfully mined icon not sure if you use a complete icon or is separated so I export both ways, let me know if that's ok or if you need anything else.

[assets.zip](https://github.com/1Hive/honey-pot/files/6185264/assets.zip)

PR #344, opened 2021-03-24
Modal state flow changes

rperez89:
This PR address this suggestion

We should add a third step to the modal for the user to allow the 1Hive protocol

We need to make some color adjustments to differentiate more the states of "Transaction is being minded..." and "Transaction successfully mined", here you can see the differentiation. Not only the icon changes, the color of the arrow too, they should be yellow when the transaction is being minded. The text should change the color too

The "Transaction being minded" changes to "Transaction is being minded..."

The "Transaction mined" text changes to "Transaction successfully mined"

The Waiting for signature step indicator, should change the grey to our grey, now it has the Aragon's one

**from #336**

### DIFF
--- a/src/components/Stepper/Step/Illustration.js
+++ b/src/components/Stepper/Step/Illustration.js
@@ -28,11 +28,10 @@ function Illustration({ status, index }) {
             display: flex;
             align-items: center;
             justify-content: center;
-            background-color: ${theme.surfaceOpened};
+            background-color: ${theme.surfaceIcon};
             height: 100%;
             width: 100%;
             border-radius: 100%;
-            color: ${theme.accentContent};
           `}
         />
       ) : (
--- a/src/components/Stepper/Step/Illustration.js
+++ b/src/components/Stepper/Step/Illustration.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import { useTheme } from '@1hive/1hive-ui'
 import {
   STEP_ERROR,
   STEP_SUCCESS,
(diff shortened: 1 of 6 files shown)


src/components/Stepper/Step/Illustration.js
```diff
@@ -9,36 +10,33 @@ import {
...
   [STEP_SUCCESS]: signRequestSuccessIllustration,
   [STEP_ERROR]: signRequestFailIllustration,
 }
 
 function Illustration({ status, index }) {
+  const theme = useTheme()
   return (
     <>
-      {/* {status === STEP_WORKING ? (
+      {status === STEP_PROMPTING ? (
```
fabriziovigevani, line 25: Might make more sense to remove this as it was before and reuse this with the div on the `StatusVisual` component and change the `renderIllustration` logic?

src/components/Stepper/Step/StatusVisual.js
```diff
@@ -209,11 +208,11 @@ function StepIllustration({ number, status, withoutFirstStep }) {
...
             justify-content: center;
-            background-color: ${theme.surfaceOpened};
+            background-color: ${theme.surfaceIcon};
             height: 100%;
             width: 100%;
             border-radius: 100%;
-            color: ${theme.accentContent};
+            color: ${theme.positiveContent};
 
             ${textStyle('title3')}
```
fabriziovigevani, line 217:
```suggestion
            ${textStyle('title3')};
```

fabriziovigevani (commented):
Awesome Ro, one small comment

Also i noticed that the animated border is a bit closer to the step circle than it should

Screen Shot 2021-03-24 at 20 12 38

src/components/Stepper/Step/Illustration.js
rperez89: i thought about that but i ended thinking that makes more sense to have it in the component in charge of rendering the illustration for each status although is not a illustration itself but act as one

rperez89:
> Awesome Ro, one small comment
> 
> Also i noticed that the animated border is a bit closer to the step circle than it should
> 
> Screen Shot 2021-03-24 at 20 12 38

Fabri not sure , this is an screenshot of the current code 
Screen Shot 2021-03-25 at 1 30 31 AM

what do you think?

src/components/Stepper/Step/Illustration.js
fabriziovigevani: Ok lets leave it as is

fabriziovigevani (approved):
LGTM!

PR #348, opened 2021-03-25
Profile logic fix

### LINKED ISSUE #317: Edit Profile screen doesn't update when changing accounts.
When in the edit profile screen the only thing that changes when changing accounts is the BrightId field. It should probably exit the edit profile screen when the account is changed so the user has to reopen it with correct field info.

fabriziovigevani:
closes #317 

- Improves animation when editing profile
- Refreshes profile when account changed


rperez89 (approved):
LGTM, so the problem was the `animated.div` ?

PR #356, opened 2021-03-31
Proposal detail timestamp

### LINKED ISSUE #257: Suggestion: Add Timestamp to detailed proposal view
When you're scrolling the list of proposals/actions there is a timestamp of when it was created. It would be nice to have that same timestamp when you're looking at the detailed proposal view

image

fabriziovigevani:
closes #257


rperez89 (approved):
LGTM!

Issue #372, opened 2021-04-08
Small issues


willjgriff:
This should read: "This amount will be locked in the collateral manager and will be used as collateral when creating actions bound by this organization's Covenant."
<img width="643" alt="Screenshot 2021-04-08 at 10 36 46" src="https://user-images.githubusercontent.com/9317743/114004273-65f38680-9856-11eb-8bd1-92add4511801.png">

I'm sure I saw this non-wrap around UI issue of the blue bar on the left fixed in Celeste. Can it be fixed here too?
<img width="380" alt="Screenshot 2021-04-08 at 10 35 49" src="https://user-images.githubusercontent.com/9317743/114004510-a18e5080-9856-11eb-82b0-9ffad997d444.png">

Clicking this proposal #0 link should actually take me to vote 0, but it goes to proposal 0 and hangs.
<img width="1277" alt="Screenshot 2021-04-08 at 10 35 16" src="https://user-images.githubusercontent.com/9317743/114004803-e6b28280-9856-11eb-8dcb-01c08d6296b5.png">

Should be: "This amount will be withdrawn from your available balance and directly credited to your enabled account." (lowercase "available" and added full stop)
<img width="631" alt="Screenshot 2021-04-08 at 10 42 33" src="https://user-images.githubusercontent.com/9317743/114005084-224d4c80-9857-11eb-8529-d079559b61fa.png">

Ideally refer to 1hive as 1Hive. I know @lkngtn isn't a fan but it looks like a mistake to me.
<img width="436" alt="Screenshot 2021-04-08 at 10 43 34" src="https://user-images.githubusercontent.com/9317743/114005246-46109280-9857-11eb-9874-57b4726089d0.png">

Lets use caps for the first letter of "Community Covenant".
Also under Action Collateral it should read "You must lock 0.1 HNY as the collateral required to create a proposal. You can manage your balance in the Collateral Manager" (added an r in your and "in the" towards the end).
Also the collateral manager link doesn't work.
<img width="604" alt="Screenshot 2021-04-08 at 11 01 05" src="https://user-images.githubusercontent.com/9317743/114008004-bd472600-9859-11eb-988b-bac338eecede.png">

The funding proposal isn't telling me what threshold I need to pass the vote. Perhaps it's because the vault is empty?
[...]

PR #374, opened 2021-04-09
QA fixes

rperez89:
fixed most of the issues on #372 

For the Link from the stake manager to the proposal when clicking the row item we need to do a little change on the connector see <https://github.com/1Hive/connect/blob/048d47908d80bc41dc349a95190c218bf48ecb4e/packages/connect-agreement/src/thegraph/parsers/staking.ts#L55> we need to change that `action.id` for `action.actionId` see <https://github.com/1Hive/connect/blob/048d47908d80bc41dc349a95190c218bf48ecb4e/packages/connect-agreement/subgraph/schema.graphql#L40>

### DIFF
--- a/src/components/Metrics.js
+++ b/src/components/Metrics.js
@@ -1,12 +1,5 @@
 import React from 'react'
-import {
-  Box,
-  // DropDown,
-  GU,
-  textStyle,
-  useLayout,
-  useTheme,
-} from '@1hive/1hive-ui'
+import { Box, GU, textStyle, useLayout, useTheme } from '@1hive/1hive-ui'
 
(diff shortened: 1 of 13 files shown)


src/components/Metrics.js
```diff
@@ -1,15 +1,14 @@
-import React, { useState, useMemo, useCallback } from 'react'
+import React from 'react'
 import {
   Box,
-  DropDown,
+  // DropDown,
```
fabriziovigevani, line 4:
```suggestion
```

fabriziovigevani (approved):
LGTM!

PR #388, opened 2021-04-12
Fix block explorer links

0xGabi:


0xGabi:
Thanks for the addition @fabriziovigevani

PR #393, opened 2021-04-14
HoneyPot Gardens, rinkeby configuration changed

famole:
In this branch Gardens will have a copy of master honeypot in order to start testing different daos in Rinkeby.

Pointed to a new DAO in rinkeby
Updated subgraph definitions for rinkeby
Defined env vars file


0xGabi:
We can close this one as it was replaced by new changes in #403

PR #397, opened 2021-04-20
Ceramic idx

famole:
Test for migrate account using IDX and Ceramic.


0xGabi:
Let's close this one as Ceramic and IDX libraries changed a lot. And keep the branch still around for reference.

PR #403, opened 2021-04-27
Dao routing with two route handlers

rperez89:


src/components/Feed/ProposalCard.js
```diff
@@ -31,14 +32,17 @@ function Card({
...
+  } = useDAO()
 
   const { below } = useViewport()
 
   const handleSelectProposal = useCallback(() => {
     const entityPath =
       proposal.type === ProposalTypes.Decision ? 'vote' : 'proposal'
-    history.push(`/${entityPath}/${proposal.number}`)
-  }, [history, proposal.number, proposal.type])
+    history.replace(`/${daoAddress}/${entityPath}/${proposal.number}`)
```
fabriziovigevani, line 44: Same comment here. 

What's more i would like to consider not having the `{daoId}/home` route and just leave it as `/{daoId}`.
This way we would add the proposal path to the current path `history.push(${history.location.pathname}/${entityPath}/${proposal.number})`

src/garden/Routes.js
```diff
@@ -0,0 +1,27 @@
...
+import { useDAO } from '../providers/Dao'
+
+export default function Routes() {
+  const { connectedDAO } = useDAO()
+  const { address: daoAddress } = connectedDAO || {}
+
+  return (
+    <Switch>
+      <Redirect exact from={`*/${daoAddress}/`} to={`${daoAddress}/home`} />
+      <Route path="*/home" component={Home} />
```
fabriziovigevani, line 18: As commented in the previous comment, what do you think of updating this to 
```suggestion
      <Route exact path={`*/${daoAddress}/`} component={Home} />
```

So children components that mutates the history, only need to use the current location pathname

src/providers/Dao.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function DAOProvider({ children }) {
+  const location = useLocation()
+  const { account } = useWallet()
+
+  // get all the daos for the dashboard
+
+  // note that i am calling this daoId because at some point we might want to also handle aragon id's and not just addresses
+  const firstParameter = parsePath(location.pathname)
```
fabriziovigevani, line 27: Now that we are using the :daoId` match, we can yse `useRouteMatch`, adding what i suggested of having `garden` as prefix path we can do

```
  const match = useRouteMatch('/garden/:daoId')
```

src/providers/Dao.js
```diff
@@ -0,0 +1,51 @@
...
+  const connectedDAO = daoList.daos.find(d =>
+    addressesEqual(firstParameter, d.address)
+  )
+
+  // somehow here we are going to get the dao token info and the user balance for the dao
+  const accountBalance = getConnectedAccountBalance(account, connectedDAO)
+  const stakeToken = {}
+
+  const DAOInfo = {
+    gardensList: daoList.daos,
```
fabriziovigevani, line 38: Lets remove this as i don't think it will be needed here

src/garden/home.js
```diff
@@ -0,0 +1,31 @@
...
+import { useDAO } from '../providers/Dao'
+
+import { DAONotFound } from '../errors'
+
+function GardenHome({ match }) {
+  const { connectedDAO } = useDAO()
+
+  if (!connectedDAO) {
+    throw new DAONotFound(match.params.daoId)
+  }
```
fabriziovigevani, line 17: Thinking that probably this should be handled in the dao provider as we will have many routes with the dao address

fabriziovigevani (commented):
Nice Ro

Left some comments about some improvements that could be made.
Let me know what do you think.


Also regarding the dao provider, not entirely convinced about exposing the `connectedDAO` property, maybe we could just return the entire dao info spread? Because i think it adds more unnecessary lines of code when components use it/.

src/garden/home.js
rperez89: ah no, i needed to remove it from the dao provider to remove the `if(path=home)` we need to let the thing go through until the specific garden tree, because the dao provider needs to be on the general gardens tree view because of the wallet component

src/components/Account/ScreenConnected.js
```diff
@@ -22,17 +23,18 @@ function AccountScreenConnected({ onClosePopover, wallet }) {
...
   const history = useHistory()
   const copy = useCopyToClipboard()
+  // const { address: daoAddress } = useDAO()
 
   const networkName = getNetworkName()
   const providerInfo = getProviderFromUseWalletId(wallet.activated)
 
   const goToProfile = useCallback(() => {
-    history.push('/profile')
+    history.replace(`/profile`)
```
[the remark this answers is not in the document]
rperez89: ah nice catch, was when i was testing some stuffs with the old structure

src/components/Feed/ProposalCard.js
rperez89: yeah make sense maybe was for that that i changed for replace

src/garden/Routes.js
rperez89: yeah!

src/providers/Dao.js
rperez89: i mean, maybe the struct does not have the right naming, but what we are returning here is the list of daos (that we should take it from the subgraph and append the dao metadata from the daoList library) and also the connected dao if there is any

fabriziovigevani: Ah yeah sounds good. Maybe we could rename to `daoList` to keep it consistent

src/garden/home.js
fabriziovigevani: Ah yeah i get what you are saying but we could bypass it checking if we got a match in the `useRouteMatch` for example 

```
export function DAOProvider({ children }) {
  const match = useRouteMatch('/garden/:daoId')

  // get all the daos for the dashboard

  // note that i am calling this daoId because at some point we might want to also handle aragon id's and not just addresses

  let connectedDao = {
    address: '',
  }

  if (match) {
    const daoAddress = match.params.daoId
    connectedDao = daoList.daos.find(d => addressesEqual(daoAddress, d.address))

    if (!connectedDao) {
      throw new DAONotFound(daoAddress)
    }
  }

  return (
    <DAOContext.Provider value={connectedDao}>{children}</DAOContext.Provider>
  )
}
```

rperez89:
@fabriziovigevani did the comments you suggested make total sense, however given that i am travelling tomorrow could't finish all what i wanted to do, we have to do some fixes on the `Profile` componente since we can't just track the activity only in one org we need to track top 3 or something like that as luke suggested, also the amount of token staked need to be splitted between each token in each org, we have to give some thought to that, see that i commented a bunch of code.
Also we need to move the collateral manager option menu to be visible just in specific Garden mode.

Since the pr is in a raw state i would suggest to not merge it, i can work on it on monday once i am back or if you want to work or need something just make commits on this one or like you want.

Sorry couldn't wrap this up

fabriziovigevani:
Lets merge this one and review #407 where we fix and accomodate the logic to be able to browse between gardens

PR #407, opened 2021-05-05
Use gardens connector

fabriziovigevani:
- Uses garden connector to fetch gardens list and merges with `dao-list` metadata.
- Implemented simple component that displays garden address and proposal count (dashboard following designs coming up in the next PR)


src/providers/Gardens.js
```diff
@@ -0,0 +1,104 @@
...
+      const gardenAddress = match.params.daoId
+      return gardens.find(d => addressesEqual(gardenAddress, d.address))
+    }
+
+    return null
+  }, [gardens, match])
+
+  if (connectedGarden) {
+    return (
+      <WithConnectedGarden connectedGarden={connectedGarden} gardens={gardens}>
```
rperez89, line 30: wondering why in this if instead of returning this new `WithConnectedGarden` we just assign the connectedGarden variable? also doing with that way we avoid possible issues trying to deconstruct when using the useGardens hook and trying to access the .connectedGarden in some place that is not present, i think we should always return it and if does not exist return null

src/utils/data-utils.js
```diff
@@ -158,5 +158,6 @@ export function getAppAddressByName(apps, appName) {
 }
 
 export function getAppByName(apps, appName) {
-  return apps?.find(app => app.name === appName)
+  const regex = new RegExp(`^${appName}.*$`)
```
rperez89, line 161: could you remind me which are. the possible app names ?

rperez89 (commented):
Looks good fabri! just some small comments

src/providers/Gardens.js
fabriziovigevani: The rationale for doing it this way was so we don't call the token balances hook when no garden is connected. If `connectedGarden` isn't present in some component that's using the `useGardens` hook, there's no real harm since from the structure that we've created, each component knows when a `connectedGarden` could or could not be present. 

That being said, in the next PR we'll lift up the Connect provider and AppState provider to the Gardens provider and encapsulate children with said providers only if there's a connected garden so components outside the `GardenHome` tree can access the garden's data. So the `WithConnectedGarden` won't be necessary

src/utils/data-utils.js
fabriziovigevani: All the ones at `environment.js`

```
const DEFAULT_CONVICTION_APP_NAME = 'disputable-conviction-voting'
const DEFAULT_VOTING_APP_NAME = 'disputable-voting'
const DEFAULT_ISSUANCE_APP_NAME = 'dynamic-issuance'
const DEFAULT_AGREEMENT_APP_NAME = 'agreement'
```

fabriziovigevani:
I'll merge this one and we can continue discussion of first comment in the next PR

PR #408, opened 2021-05-06
Gardens dashboard

fabriziovigevani:
- Implements basic gardens layout
- Uses specific garden metadata to display loco, icon and token symbol|name, if not present uses default logo and icons.
- Lifts Connect provider and AppState provider to Gardens provider so components outside `GardenHome`(e.g `BalanceModule`) tree have garden's data available.
- Created utility function to build a path relative to current connected garden

Still need to figure out what metric should we use for members count

Screen Shot 2021-05-06 at 19 43 45


src/components/GardensDashboard.js
```diff
@@ -1,27 +1,122 @@
...
+        <GardenCard key={garden.id} garden={garden} />
       ))}
     </div>
   )
 }
 
+function GardenCard({ garden }) {
+  const theme = useTheme()
+  const history = useHistory()
+  const handleSelectGarden = useCallback(() => {
```
rperez89, line 31: not sure if we want to keep the buildGardenPath or not but if we decide to use it we should use it here also right?

src/components/GardensDashboard.js
```diff
@@ -1,27 +1,122 @@
...
+    >
+      <div
+        css={`
+          height: 72px;
+          display: flex;
+          align-items: center;
+          justify-content: center;
+        `}
+      >
+        <img src={garden.logo || defaultGardenSvg} alt="" />
```
rperez89, line 60: lets create a memo with ` const cardData = useMemo(() => {logo: garden.logo || defaultGardenSvg, name: garden.name || shortenAddress(garden.id) ........}, [gardens])`

src/hooks/useProposals.js
```diff
@@ -279,11 +280,11 @@ function processProposal(
     alpha
   )
 
-  const userStakedConviction = userConviction.div(maxConviction)
+  const userStakedConviction = safeDivBN(userConviction, maxConviction)
 
-  const stakedConviction = currentConviction.div(maxConviction)
+  const stakedConviction = safeDivBN(currentConviction, maxConviction)
```
rperez89, line 285: oh did you noticed some error?

rperez89 (commented):
Awesome work fabri! some comments and questions

src/utils/routing-utils.js
```diff
@@ -0,0 +1,6 @@
+const GARDEN_PATH_REGEX = /(.*?garden\/0x[0-9a-zA-Z]{40}).*/ // TODO: Update when aragonIds supported
+
+export function buildGardenPath(location, relativePath) {
```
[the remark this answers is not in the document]
fabriziovigevani: Because if for example we are in `*/garden/0x../collateral` and we want to redirect to a proposal `*/garden/0x../proposal/0` we can't just push to the current location, we need to remove the replace the relative path with the new one

src/components/GardensDashboard.js
fabriziovigevani: Not in this case because it assumes the current location is a garden

src/components/GardensDashboard.js
fabriziovigevani: What benefit do you see in this case? I think having a `useMemo` would result in more expensive computation than the simple operations we are performing.

rperez89: well discard the use memo, and we can just have the variable instanced before the html so we don't have the `||` operator and we use the variable directly

rperez89:
@fabriziovigevani merging this to move forward

PR #413, opened 2021-05-10
Use user stakes

fabriziovigevani:
- Uses user's data following changes done in the connector at https://github.com/1Hive/gardens/pull/104
- Fixes supporter distribution display on Proposal details
- Profile: Fixes Activity and Tokens staked distribution
- Profile: Updates links to garden specific proposals

### DIFF
--- a/src/components/Profile/StakingTokens.js
+++ b/src/components/Profile/StakingTokens.js
@@ -16,8 +16,7 @@ function displayedStakes(stakes, total) {
     }
   ).map((stake, index) => ({
     item: {
-      organization:
-        stake.index === -1 ? null : stakes[stake.index].organization,
+      gardenId: stake.index === -1 ? null : stakes[stake.index].gardenId,
       proposalId: stake.index === -1 ? null : stakes[stake.index].proposalId,
       proposalName:
         stake.index === -1 ? 'Others' : stakes[stake.index].proposalName,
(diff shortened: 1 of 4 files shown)


src/components/Profile/StakingTokens.js
```diff
@@ -16,6 +16,8 @@ function displayedStakes(stakes, total) {
     }
   ).map((stake, index) => ({
     item: {
+      organization:
```
rperez89, line 19: let's rename this to garden for consistency

src/hooks/useAccountTotalStaked.js
```diff
@@ -1,10 +1,10 @@
 import { useMemo } from 'react'
-import { useAccountStakes } from './useStakes'
+import { useAccountStakesByGarden } from './useStakes'
 
 import BigNumber from '../lib/bigNumber'
 
 export default function useAccountTotalStaked(account) {
-  const stakes = useAccountStakes(account)
+  const stakes = useAccountStakesByGarden(account)
```
rperez89, line 7: we don't need to pass the account  since we can use the useWallet inside the hook, there are several places were we are calling it with the account

rperez89 (commented):
awesome fabri!

src/hooks/useAccountTotalStaked.js
rperez89: ah yeah i.e the profile that sometimes we get the address from the url rihgt? if that the case makes sense yes

src/utils/garden-utils.js
```diff
@@ -0,0 +1,7 @@
+import daoList from '@1hive/gardens-dao-list'
+import { addressesEqual } from './web3-utils'
+
+export function getGardenLabel(address) {
```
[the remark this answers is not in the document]
fabriziovigevani: Initially i used `name` but the changed it to `label` because the function returns the `address` of the garden if there's no `name` associated with the garden. And i don't think `title` is accurate in this case. Why do you think `label` is not the right one here?

PR #421, opened 2021-05-13
withdraw fix

rperez89:
Fix for <https://discord.com/channels/698287700834517064/770282095758802964/842096378145406996>

### DIFF
--- a/src/components/ProposalDetail/ProposalActions.js
+++ b/src/components/ProposalDetail/ProposalActions.js
@@ -33,17 +33,17 @@ function ProposalActions({
   const didIStake = myStake?.amount.gt(0)
 
   const mode = useMemo(() => {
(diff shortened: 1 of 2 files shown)


fabriziovigevani (approved):
LGTM!

When testing it, i noticed that after you withdraw your stake, the `Execute` button appears for some reason (since probably the conviction is still above the threshold)

We should update how we calculate the `mode` on ProposalActions to: 

```
  const mode = useMemo(() => {
    if (hasEnded) {
      if (didIStake) {
         return 'withdraw' 
      }
      return null
    }
    if (currentConviction.gte(threshold)) {
      return 'execute'
    }
    if (didIStake) {
      return 'update'
    }
    return 'support'
  }, [currentConviction, didIStake, hasEnded, threshold])
```

PR #434, opened 2021-05-19
Fix tokens

fabriziovigevani:
- Updates Wrap Unrwap token icons (were a bit stretched)
- Header: Uses wrapable token name on BYOT garden
- Metrics: Uses wrapable token name on BYOT garden
- Decisions: Uses voting token as opposed to stake token


src/components/Metrics.js
```diff
@@ -16,14 +16,16 @@ const Metrics = React.memo(function Metrics({
   const { layoutName } = useLayout()
   const compactMode = layoutName === 'small'
   const {
-    connectedGarden: { token },
+    connectedGarden: { wrappableToken, token },
   } = useGardens()
```
rperez89, line 20: taking a look at the way that you implemented makes me wonder if we could identify on the Gardens provider if the dao is BYOT or normal and have new attr called `metricsToken` or `accountableToken` or whatever we want to call it and return it in the hook, that way we avoid `const metricsToken = wrappableToken || token` that we are doing on each component

rperez89 (commented):
Looks good! just one small comment to improve a bit the logic

src/components/Metrics.js
fabriziovigevani: Yeah i think we can do that when we do the general refactor of the logic which i was intending to do after reviewing the folder restructure PR.

fabriziovigevani:
Closing in favor of #446

PR #442, opened 2021-05-21
UX: add intercom integration

0xGabi:
Setup intercom using the `react-use-intercom` hook.

Screen Shot 2021-05-21 at 6 34 08 PM


fabriziovigevani (approved):
🚀

PR #444, opened 2021-05-21
Initial support for WalletConnect

0xGabi:
Screen Shot 2021-05-28 at 03 06 01
Screen Shot 2021-05-28 at 03 06 06


src/ethereum-providers/icons/walletConnect.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rperez89, line 1: let's optimize this file, not sure if you know how to do it but in. case you don't you need to get the library svgo and run `svgo .... file path`

src/providers/Wallet.js
```diff
@@ -40,8 +40,9 @@ function WalletAugmented({ children }) {
 
 function WalletProvider({ children }) {
   const chainId = getDefaultChain()
+  const network = getNetwork(chainId)
 
-  const connectors = getUseWalletConnectors()
+  const connectors = getUseWalletConnectors(network)
```
rperez89, line 45: wondering if we can do something similar to swapr, because in my opinion doesn't make to much sense to me change a library that is not being maintained anymore for another library that neither is being maintaned, but if we can tune it in some way like the swapr one would be pretty cool, having the ability to choose the network and change between gardens deployments without the need of use another endpoint 

Screen Shot 2021-05-21 at 9 04 52 PM

Screen Shot 2021-05-21 at 9 07 44 PM

rperez89 (commented):
cool gaby some comments about if makes sense to just change it without having any benefit

src/providers/Wallet.js
rperez89: Yo can discard this comment

0xGabi:
Replaced by https://github.com/1Hive/honey-pot/pull/672

PR #446, opened 2021-05-22
Reestructure garden logic

### LINKED ISSUE #439: Re think AppState logic
Since the app is growing too much we need to give a thought to the AppState structure since is getting pretty convoluted

fabriziovigevani:
closes #439 

- Removes `stakeToken` and `requestToken` from garden state object first level attributes and updates components that use them to use `config.conviction`
- Same as previous item but for `totalStaked` 
- Moves `effectiveSupply` to `config.conviction` (note that in this case this attribute didn't previously existed in the `config.conviction` object) 
- Decisions: Updates to use `config.voting.token` (Previously was using `config.conviction.stakeToken`)
- Renamed `isLoading` for `loading` everywhere to keep consistency
- Created hook `useTokens` that returns an object with `token` and `wrappableToken` (if present) with the following format: 
```
{
  accountBalance, // connected account's balance
  data // token metadata
  totalSupply  // token's total supply
}
```
- Renamed `vaultBalance` for `commonPool`


rperez89 (approved):
Nice job fabri!!

PR #461, opened 2021-05-24
Fixes

### LINKED ISSUE #447: `Create proposal` button is not popping up modal
After signing covenant, depositing collateral, the `Create proposal` button redirects to the garden home page but doesn't pop up the Proposal creation modal

### LINKED ISSUE #448: Challenge action:  Approve token amounts on necessary tokens
When challenging a proposal, it asumes colateral Requirement token and disputeFee token is the same. Which is not true for gardens other than 1Hive.

### LINKED ISSUE #450: Challenge proposal: when using BYOT the token logic is not correct
When handling the Challenge action the logic should consider if the Garden has a `WrappableToken` and in that case use that on as the OGT instead. Now we are considering the OGT but the balances displayed are for the GDN token.

Screen Shot 2021-05-22 at 3 06 00 PM

In the following picture, I have OGT balances and anyways the modal show me I don't.

Screen Shot 2021-05-22 at 3 04 51 PM

fabriziovigevani:
closes #448 , closes #450, closes #447

- Fixes create proposal modal not popping when being redirected from collateral manager
- Challenge actions: 2 Approves (one for collateral amount, one for dispute fee amount) since collateral token and dispute fee token can be different. If tokens are the same, then one approve for the added amount.
- useDisputeFees: Returns amount converted in BN so that components don't need to handle this.   
- Challenge requirements: Uses dispute fee token data and account balance (previously using collateral token data)
- Raise Dispute requirements: Uses dispute fee token data and account balance (previously using collateral token data)

### DIFF
--- a/src/components/Garden/ActionCollateral.js
+++ b/src/components/Garden/ActionCollateral.js
@@ -7,6 +7,8 @@ import lockIconSvg from '@assets/icon-lock.svg'
 
 function ActionCollateral({ proposal }) {
   const { collateralRequirement } = proposal
+  const tokenIcon = getTokenIconBySymbol(collateralRequirement.tokenSymbol)
+
   return (
     <div
       css={`
@@ -15,7 +17,7 @@ function ActionCollateral({ proposal }) {
       `}
     >
       <img
-        src={getTokenIconBySymbol(collateralRequirement.tokenSymbol)} // TODO: Use deefault-token-list
+        src={tokenIcon} // TODO: Use deefault-token-list
         alt=""
         height="24"
         width="24"
--- a/src/components/Garden/ActionCollateral.js
+++ b/src/components/Garden/ActionCollateral.js
@@ -15,7 +15,7 @@ function ActionCollateral({ proposal }) {
       `}
     >
       <img
-        src={getTokenIconBySymbol(collateralRequirement.tokenSymbol)}
+        src={getTokenIconBySymbol(collateralRequirement.tokenSymbol)} // TODO: Use deefault-token-list
         alt=""
         height="24"
         width="24"
--- a/src/components/Garden/ActionCollateral.js
+++ b/src/components/Garden/ActionCollateral.js
@@ -2,7 +2,7 @@ import React from 'react'
 import { GU } from '@1hive/1hive-ui'
 
 import { formatTokenAmount } from '@utils/token-utils'
-import honeyIconSvg from '@assets/honey.svg'
+import { getTokenIconBySymbol } from '../../utils/token-utils'
 import lockIconSvg from '@assets/icon-lock.svg'
 
 function ActionCollateral({ proposal }) {
@@ -15,12 +15,12 @@ function ActionCollateral({ proposal }) {
       `}
     >
       <img
-        src={honeyIconSvg}
+        src={getTokenIconBySymbol(collateralRequirement.tokenSymbol)}
         alt=""
-        height="28"
-        width="28"
+        height="24"
+        width="24"
         css={`
-          margin-right: ${0.5 * GU}px;
+          margin-right: ${1 * GU}px;
         `}
       />
       <div
--- a/src/components/Garden/Home.js
+++ b/src/components/Garden/Home.js
@@ -45,13 +45,6 @@ const Home = React.memo(function Home() {
     setFilterSidlerVisible(visible => !visible)
   }, [])
 
-  useEffect(() => {
-    // Components that redirect to create a proposal will do so through "garden/${gardenId}/create" url
-    if (account && history.location.pathname.includes('create')) {
-      setModalVisible(true)
-    }
-  }, [account, history])
-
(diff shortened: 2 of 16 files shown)


src/components/Garden/Home.js
```diff
@@ -77,6 +70,15 @@ const Home = React.memo(function Home() {
     handleShowModal('unwrap')
   }, [handleShowModal])
 
+  useEffect(() => {
+    console.log('running')
```
rperez89, line 74: let's remove this console log

src/components/Garden/Home.js
```diff
@@ -77,6 +70,15 @@ const Home = React.memo(function Home() {
...
+    // Components that redirect to create a proposal will do so through "garden/${gardenId}/create" url
+    if (account && history.location.pathname.includes('create')) {
+      console.log('entered')
```
rperez89, line 77: also this one

src/components/Garden/ModalFlows/ChallengeProposalScreens/ChallengeProposalScreens.js
```diff
@@ -7,51 +7,68 @@ import { useDisputeFees } from '@hooks/useDispute'
 import { useAgreement } from '@hooks/useAgreement'
 import BigNumber from '@lib/bigNumber'
 import { toDecimals } from '@utils/math-utils'
+import { addressesEqual } from '@1hive/connect-core'
```
rperez89, line 10: for consistency let's use or the one in utils or the one on the 1hive-ui

src/components/Garden/ModalFlows/ChallengeProposalScreens/ChallengeProposalScreens.js
```diff
@@ -7,51 +7,68 @@ import { useDisputeFees } from '@hooks/useDispute'
...
+  const getTransactions = useCallback(
+    async (
+      onComplete,
+      actionId,
+      settlementOffer,
+      challengerFinishedEvidence,
+      context
+    ) => {
+      const collateralToken = proposal.collateralRequirement.tokenId
+      const collateralAmount = proposal.collateralRequirement.challengeAmount
```
rperez89, line 56:
```suggestion
      const {challengeAmount: collateralAmount, tokenId: collateralToken} = proposal.collateralRequirement
```

src/components/Garden/ModalFlows/ChallengeProposalScreens/ChallengeProposalScreens.js
```diff
@@ -7,51 +7,68 @@ import { useDisputeFees } from '@hooks/useDispute'
...
+
+      const disputeFeeToken = disputeFees.token
+      const disputeFeeAmount = disputeFees.amount
```
rperez89, line 59:
```suggestion
      const {amount: disputeFeeAmount, token: disputeFeeToken}  = disputeFees
```

src/components/Garden/ModalFlows/ChallengeProposalScreens/ChallengeProposalScreens.js
```diff
@@ -71,7 +93,7 @@ function ChallengeProposalScreens({ agreementActions, proposal }) {
         content: (
           <ChallengeRequirements
             agreement={agreement}
-            accountBalance={token.accountBalance}
+            accountBalance={(wrappableToken || token).accountBalance}
```
rperez89: let's create a collateralToken var outside

rperez89 (commented):
looks good! some comments

src/components/Garden/ActionCollateral.js
```diff
@@ -15,12 +15,12 @@ function ActionCollateral({ proposal }) {
       `}
     >
       <img
-        src={honeyIconSvg}
+        src={getTokenIconBySymbol(collateralRequirement.tokenSymbol)} // TODO: Use deefault-token-list
```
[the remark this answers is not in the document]
fabriziovigevani: I think that's fine though since it would just show the default token icon. I don't have strong feelings with keeping it this way or using the `default-token-list` but i think thos way we don't maintain duplicated metadata.

Nevertheless it should be noted that if we go with our current implementation we will need to update this function (`getTokenIconBySymbol `), since it just returns the icon from a mapping of harcoded symbol to icon, it's not a general solution.

src/components/Garden/ModalFlows/ChallengeProposalScreens/ChallengeProposalScreens.js
fabriziovigevani: In this case i would prefer to keep is as since the suggested code takes 4 lines when auto-linted

src/components/Garden/ActionCollateral.js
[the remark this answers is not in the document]
rperez89: > maintain duplicated

yeah in any case we can fetch it from the default token list and fallback to the dao list in case not found it

PR #463, opened 2021-05-26
Fix various bugs

fabriziovigevani:


src/providers/Gardens.js
```diff
@@ -78,10 +78,18 @@ function mergeGardenMetadata(garden) {
     ...garden.token,
     ...metadata.token,
   }
+  const wrappableToken = garden.wrappableToken
```
rperez89, line 81: let's use the one line syntax `garden.wrappableToken  && {  ...garden.wrappableToken,        ...metadata.wrappableToken  }`

PR #469, opened 2021-05-26
Landing Banner

fabriziovigevani:
Large
Screen Shot 2021-05-26 at 11 54 41

Medium
Screen Shot 2021-05-26 at 11 56 04

Mobile
Screen Shot 2021-05-26 at 12 46 09


src/components/GardensDashboard.js
```diff
@@ -1,124 +1,24 @@
...
+import React from 'react'
+import styled from 'styled-components'
+import GardensList from './GardensList'
+import LandingBanner from './LandingBanner'
+import { useNodeHeight } from '@hooks/useNodeHeight'
 
 function GardensDashboard() {
-  // TODO :  add loading component
-  const { gardens } = useGardens()
+  const [height, ref] = useNodeHeight()
```
fabriziovigevani, line 8: This hook tracks the landing Banner container's height so we can dynamically set the gardens list container margin.

This is because the banner is absolute positioned (since it's inside the `Layout` component which set's left and right margins resulting in the banner not having the full window width otherwise) so we need to place the gardens list container perfectly below the banner.

src/components/GardensDashboard.js
```diff
@@ -1,124 +1,24 @@
...
-    <div
-      onClick={handleSelectGarden}
-      css={`
-        padding: ${5 * GU}px ${4 * GU}px ${3 * GU}px ${4 * GU}px;
-        background: ${theme.surface};
-        border: 1px solid ${theme.border};
-        border-radius: ${2.5 * GU}px;
-        box-shadow: 0px 1px 2px rgba(0, 0, 0, 0.15);
-        cursor: pointer;
+const DynamicDiv = styled.div.attrs(props => ({
```
fabriziovigevani, line 20: This is a suggestion done by styled-components.
Setting the margin through `css={``}` directly resulted in styled components creating a class for each value

rperez89 (approved):
Awesome work fabri, as you explained the implementation in a call feel free to merge!!!

PR #473, opened 2021-05-26
Execute  proposal status

### LINKED ISSUE #462: Proposal: Showing available for execution when proposal is challenged
When a proposal is challenged and it reached sufficient conviction, the execute proposal button is displayed when it shouldn't.

rperez89:
This closes #462

### DIFF
--- a/src/components/Garden/ProposalDetail/ProposalActions.js
+++ b/src/components/Garden/ProposalDetail/ProposalActions.js
@@ -41,13 +41,16 @@ function ProposalActions({
       return null
     }
     if (currentConviction.gte(threshold)) {
-      return 'execute'
+      if (proposal.statusData.open) {
+        return 'execute'
+      }
+      return null
     }
     if (didIStake) {
       return 'update'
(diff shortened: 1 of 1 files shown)


src/components/Garden/ProposalDetail/ProposalActions.js
```diff
@@ -41,13 +41,16 @@ function ProposalActions({
       return null
     }
     if (currentConviction.gte(threshold)) {
-      return 'execute'
+      if (proposal.statusData.open) {
+        return 'execute'
+      }
+      return null
```
fabriziovigevani, line 47: Probably we don't want to hide actions when the proposal is challenged or disputed.

Let's say i staked and the proposal got challenged and it reached sufficient conviction but i want to update my support. 

```suggestion
```

rperez89: good catch!

fabriziovigevani (approved):

PR #486, opened 2021-05-27
Fix generic error screen

fabriziovigevani:
- Reorganized nested providers so that GlobalErrorHandler is a child of the `Main` component
- Header: Updated the logo link depending on if there's a connected garden or not

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -1,10 +1,13 @@
 import React, { useMemo } from 'react'
+import { useHistory } from 'react-router'
 import { GU, Link, useTheme, useViewport } from '@1hive/1hive-ui'
 import AccountModule from '../Account/AccountModule'
 import BalanceModule from '../BalanceModule'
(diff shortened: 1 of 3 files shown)


src/components/Header/Header.js
```diff
@@ -1,10 +1,13 @@
 import React, { useMemo } from 'react'
```
0xGabi, line 1: I'm not sure what I should try to reproduce here. Is the intention to alway get to the particular Garden's home? It seems it always take me to the Dashboard.

0xGabi (commented):
Left a comment. Other than that looking good!

src/components/Header/Header.js
fabriziovigevani: Yes, if you are in a specific garden then clicking on the logo should direct you to that garden's home.

Before it would take you to the gardens list view.

0xGabi (approved):
🙌

PR #489, opened 2021-05-28
Claim collateral: Update copy when account is challenger

fabriziovigevani:
Screen Shot 2021-05-28 at 01 30 30
Screen Shot 2021-05-28 at 01 30 11


0xGabi (approved):
Good catch! 💪

PR #490, opened 2021-05-28
Fix footer bottom

fabriziovigevani:
Before
Screen Shot 2021-05-28 at 01 24 05

After
Screen Shot 2021-05-28 at 01 24 27


0xGabi (approved):
LGTM! I change the base cause it was building on #489

PR #493, opened 2021-05-28
Use default provider

fabriziovigevani:
Uses as provider the `defaultProvider` from ethers only if no `ETH_NODE` is set on startup. This is because most of the backend services accepted by `defaultProvider` don't support xdai as far as i could tell. 

It's also nice to have this possibility so that when developing you only need to pass your personal eth endpoint instead of all your backend services api keys (if any).

Note that when no api keys are provided `getDefaultProvider` uses some default keys which are used amongst all devs.

TODO: Add pocket network

### DIFF
--- a/src/utils/web3-utils.js
+++ b/src/utils/web3-utils.js
@@ -9,7 +9,7 @@ function getBackendServicesKeys() {
   return {
     alchemy: env('ALCHEMY_API_KEY'),
     etherscan: env('ETHERSCAN_API_KEY'),
-    infura: env('INFURA_API_KEY') || null,
+    infura: env('INFURA_API_KEY'),
     pocket: env('POCKET_API_KEY'),
   }
 }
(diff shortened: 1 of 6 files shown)


src/utils/web3-utils.js
```diff
@@ -1,9 +1,28 @@
...
+  return {
+    alchemy: env('ALCHEMY_API_KEY'),
+    etherscan: env('ETHERSCAN_API_KEY'),
+    infura: env('INFURA_API_KEY'),
+    pocket: env('POCKET_API_KEY'),
+  }
+}
+
+export function getDefaultProvider() {
+  const type = getNetworkType()
```
0xGabi, line 18: I think the current solution is fine. In a future version, we should provide the whole `Network` object with the specific `ENS` registry we use per network. 

This will not become a problem until we want to support AragonID.

0xGabi (approved):
Great work. This is going to be a great improvement going forward 🙌

src/utils/web3-utils.js
fabriziovigevani: Good point

PR #497, opened 2021-05-31
Collateral manager: Improve deposit flow

### LINKED ISSUE #449: Deposit collateral: Promp to deposit when user directed from the create proposal modal.
It would be nice to pop up the Deposit collateral modal when the user is coming from the create proposal modal.
So that the user doesn't need to click the button.

fabriziovigevani:
closes #449 

Was wondering if we should apply the same behavior to the Allow garden action since it straight up prompts you to sign the tx? 

When going to the collateral screen from the proposal requirements modal
Screen Shot 2021-05-31 at 15 47 57

The new route is `collateral/deposit` which if detected it will open up the modal after page load.
Screen Shot 2021-05-31 at 15 48 06


rperez89 (approved):
awesome!

PR #506, opened 2021-06-16
Onboarding backbone

fabriziovigevani:
closes https://github.com/1Hive/gardens/issues/154

Includes: 

- Basic component for all screens involved with navigation.
- Progress sidebar
- Onboarding provider with basic logic (next step, back step, updating config data)
  It should be noted when implementing a screen that when submitting the form with all the inputs the data should be saved in the provider using `onConfigChange` like so: 

	 ```
	 onConfigChange(key, data)
	 ```
	  
	   where key is the respective key set in the default config (see Onboarding.js)

### DIFF
--- a/src/components/Onboarding/Navigation.js
+++ b/src/components/Onboarding/Navigation.js
@@ -45,7 +45,7 @@ const Navigation = React.forwardRef(function Navigation(
       <Button
         ref={nextRef}
         disabled={!nextEnabled}
-        label="Next"
+        label={nextLabel}
         mode="strong"
         onClick={onNext}
         type="submit"
--- a/src/components/Onboarding/Navigation.js
+++ b/src/components/Onboarding/Navigation.js
@@ -1,0 +1,76 @@
+import React, { useRef, useImperativeHandle } from 'react'
+import PropTypes from 'prop-types'
+import { Button, IconArrowLeft, GU, useTheme } from '@1hive/1hive-ui'
+
+const Navigation = React.forwardRef(function Navigation(
+  { backEnabled, backLabel, nextEnabled, nextLabel, onBack, onNext },
+  ref
+) {
+  const theme = useTheme()
+
+  const nextRef = useRef()
+
+  useImperativeHandle(
+    ref,
+    () => ({
+      focusNext: () => {
+        if (nextRef.current) {
+          nextRef.current.focus()
+        }
+      },
+    }),
+    []
+  )
+
+  return (
+    <div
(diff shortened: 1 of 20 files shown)


src/providers/Onboarding.js
```diff
@@ -0,0 +1,113 @@
+import React, { useCallback, useContext, useState } from 'react'
+import PropTypes from 'prop-types'
+import { Screens } from '@components/Onboarding/Screens/config'
+
+const OnboardingContext = React.createContext()
+
+const stepsLength = Screens.length
+
+const DEFAULT_CONFIG = {
```
fabriziovigevani, line 9: Default values, we still need to define what defaults we'll use for some of them

src/components/Onboarding/Screens/index.js
```diff
@@ -0,0 +1,72 @@
...
+              left: 0;
+              right: 0;
+            `}
+          >
+            <div
+              css={`
+                margin-bottom: ${2 * GU}px;
+              `}
+            >
+              <Screen title={title} />
```
0xGabi, line 63: Wonder how we use the prop `title` here. As I notice we did not use `title` on screens. It is because we will be using it once we finish them?

src/providers/Onboarding.js
0xGabi: Also, we should agree on the naming for each to align with the latest discussions.

0xGabi (approved):
Awesome work @fabriziovigevani. Simple and powerful logic 💪

I omit the review for the `react-spring` sections. I still need to have a better understanding of it 😅

I learned a lot by reviewing it!

src/components/Onboarding/Navigation.js
```diff
@@ -0,0 +1,76 @@
...
+          <IconArrowLeft
+            css={`
+              color: ${theme.accent};
+            `}
+          />
+        }
+        label={backLabel}
+        onClick={onBack}
+      />
+      <Button
```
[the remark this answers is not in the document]
fabriziovigevani: Ah yes i removed it temporarily and forgot to pass the screen title as `nextLabel`.
Regarding the arrow icon for the next button i wold omit it as the label is the next screen title and would look pretty long if we use it.

src/components/Onboarding/Screens/index.js
fabriziovigevani: Ah yes good catch 👍

0xGabi (approved):

PR #507, opened 2021-06-17
Onboarding: Garden type selection screen

fabriziovigevani:
closes https://github.com/1Hive/gardens/issues/153

TODO: 

- [ ] Update garden type icons (icons in screen are placeholders)
- [ ] Define if the garden type paragraph makes sense or it should have more information.

**Preview**
Screen Shot 2021-06-17 at 15 50 08

### DIFF
--- a/src/components/Onboarding/Screens/GardenTypeSelector.js
+++ b/src/components/Onboarding/Screens/GardenTypeSelector.js
@@ -1,20 +1,144 @@
-import React from 'react'
+import React, { useCallback, useState } from 'react'
+import { GU, textStyle, useTheme } from '@1hive/1hive-ui'
 import { useOnboardingState } from '@providers/Onboarding'
 import Navigation from '../Navigation'
 
+import { BYOT_TYPE, NATIVE_TYPE } from '../constants'
+import defaultGardenLogo from '@assets/defaultGardenLogo.png'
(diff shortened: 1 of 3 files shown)


src/components/Onboarding/Screens/GardenTypeSelector.js
```diff
@@ -1,20 +1,144 @@
...
+      </div>
+      <div
+        css={`
+          display: flex;
+          align-items: center;
+          justify-content: space-around;
+          margin-bottom: ${6 * GU}px;
+        `}
+      >
+        <Card
```
0xGabi, line 62: Wonder what you think of also add the `handleNext` handler to the `onSelect` prop of each `Card`?

0xGabi (approved):
Well done! The logic was straight forward with the backbone in place 💪

src/components/Onboarding/Screens/GardenTypeSelector.js
fabriziovigevani: Ah yes i thought about it but wasn't sure, let's do it 👍

0xGabi (approved):
Let's fix the merge conflicts and we can merge it as well. Let's add a couple of new issues to address the TODOs

PR #511, opened 2021-06-20
Onboarding: Add TokenSettings Screen

pavloMandryk:
This PR contains the TokensSettings screen for the Onboarding, it consists in:

- A folder with icons needed in the Header.
- The Header component used in every wizard screen.
- The KnownAppBadge component to display the icons in the Header.
- The TokenSettings screen.


0xGabi (changes requested):
@pavloMandryk I left a few comments and suggestions. We are going in the right direction with the changes. Though there is one big caveat we are missing. 

The screen is dependent on the type of Garden the user chooses on the type selection step: https://github.com/1Hive/honey-pot/pull/507

The current changes you implemented fully cover the `NATIVE` case.

We now need to support the `BYOT` case. And showing the specific screen depending on the type value that was set up in the config data: https://github.com/1Hive/honey-pot/blob/onboarding-garden-type/src/providers/Onboarding.js#L18


The requirements for the BYOT type of Garden were specified here: https://www.notion.so/Creation-Wizard-1625333dd1a148e78a1938d26d77abf9

Let me know if I can further clarify if needed.

PR #515, opened 2021-06-24
Onboarding: Add Voting Settings screen

PJColombo:
This PR implements the `<VotingSettings/>` screen for the gardens onboarding.

It also includes a series of common components inside the `Onboarding/kit` folder that can be reused in other onboarding screens.

pr

### DIFF
--- a/src/components/Onboarding/Screens/Apps/VotingSettings.js
+++ b/src/components/Onboarding/Screens/Apps/VotingSettings.js
@@ -1,19 +1,265 @@
-import React from 'react'
+import React, {
+  Fragment,
+  useCallback,
+  useReducer,
+  useRef,
+  useState,
+} from 'react'
+import { GU, Help } from '@1hive/1hive-ui'
 import { useOnboardingState } from '@providers/Onboarding'
 import Navigation from '../../Navigation'
+import {
+  TimeParameterPanel,
+  Header,
+  PercentageField,
+  Carousel,
+} from '../../kit'
+import { DAY_IN_SECONDS } from '@/utils/kit-utils'
 
-function VotingSettings({ title }) {
-  const { onBack, onNext, step, steps } = useOnboardingState()
+const DEFAULT_SUPPORT = 50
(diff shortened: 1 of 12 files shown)


PJColombo:
Hey @fabriziovigevani, it seems you can't change the base branch of a closed PR so I've created a new one

src/components/Onboarding/Screens/Apps/VotingSettings.js
```diff
@@ -1,19 +1,265 @@
...
+    },
+    updateField,
+  ] = useReducer(reduceFields, {
+    support: DEFAULT_SUPPORT,
+    quorum: DEFAULT_MIN_ACCEPTANCE_QUORUM,
+    duration: DEFAULT_DURATION,
+    delegatedVotingPeriod: DEFAULT_DELEGATED_VOTING_PERIOD,
+    quiteEndingPeriod: DEFAULT_QUITE_ENDING_PERIOD,
+    quiteEndingExtensionPeriod: DEFAULT_QUITE_ENDING_EXTENSION_PERIOD,
+    executionDelayPeriod: DEFAULT_EXECUTION_DELAY_PERIOD,
```
fabriziovigevani, line 92: We should use the `config.voting` that comes for the onboarding state provider (and set all this defaults there). Otherwise when coming back to this screen, the values would be reseted. Also was wondering if we should use the same naming convention for the attributes

fabriziovigevani (commented):
Awesome work @PJColombo!

Really liked the Carousel.
Don't have strong comments other than the below.

src/components/Onboarding/Screens/Apps/VotingSettings.js
PJColombo: Yes, we should use the same naming convention.

PJColombo:
@fabriziovigevani  I added some validation logic that was missing and fixed  a small typo on the delegated voting period panel

fabriziovigevani (approved):
LGTM!

PR #517, opened 2021-06-26
Onboarding: Add TokensSettings Screen

pavloMandryk:
This PR contains the TokensSettings screen for the Onboarding, it consists in:

- A folder with the Tokens icon for the Header.
- The Header component used in every wizard screen.
- The KnownAppBadge component to display the icons in the Header.
- The TokenSettings main screen.
     - The TokensSettingsNative screen that displays when the user selects "Native" when choosing Garden Type.
     - The TokensSettingsBYOT screen that displays when the user selects "BYOT" when choosing Garden Type.

- Adding changes to the useToken hook to fetch the token symbol, set loading only when an address is provided and treat an exception.

### DIFF
--- a/src/components/Onboarding/Screens/Apps/TokensSettings.js
+++ b/src/components/Onboarding/Screens/Apps/TokensSettings.js
@@ -1,21 +1,16 @@
 import React from 'react'
-import { useOnboardingState } from '@providers/Onboarding'
-import Navigation from '../../Navigation'
+import { useOnboardingState } from '../../../../providers/Onboarding'
+import { NATIVE_TYPE, BYOT_TYPE } from '../../constants'
+import TokensSettingsNative from './TokensSettingsNative'
+import TokensSettingsBYOT from './TokensSettingsBYOT'
 
-function TokensSettings({ title }) {
-  const { onBack, onNext, step, steps } = useOnboardingState()
-
+function TokensSettings() {
+  const { config } = useOnboardingState()
   return (
-    <div>
-      {title}
-      <Navigation
-        backEnabled
-        nextEnabled
-        nextLabel={`Next: ${steps[step + 1].title}`}
-        onBack={onBack}
-        onNext={onNext}
-      />
-    </div>
+    <>
+      {config.garden.type === NATIVE_TYPE && <TokensSettingsNative />}
+      {config.garden.type === BYOT_TYPE && <TokensSettingsBYOT />}
+    </>
   )
 }
 
--- a/src/components/Onboarding/Screens/Apps/TokensSettingsBYOT.js
+++ b/src/components/Onboarding/Screens/Apps/TokensSettingsBYOT.js
@@ -1,0 +1,282 @@
+import React, { useCallback, useEffect, useState } from 'react'
+import {
+  Field,
+  GU,
+  Help,
+  Info,
+  IconCheck,
+  IconCross,
+  isAddress,
+  LoadingRing,
+  TextInput,
+  theme,
+} from '@1hive/1hive-ui'
+import Navigation from '../../Navigation'
+import { useOnboardingState } from '@providers/Onboarding'
+import { useTokenData } from '@hooks/useToken'
+import Header from '../../kit/Header'
+
+function useFieldsLayout() {
+  return `
+    display: grid;
+    grid-template-columns: auto ${21 * GU}px;
+    grid-column-gap: ${1.5 * GU}px;
+  `
+}
+function validationError(tokenAddress, tokenName, tokenSymbol) {
+  if (!tokenAddress) {
+    return 'Please, provide a token address'
+  }
+  if (!isAddress(tokenAddress)) {
+    return 'The token address you provided is invalid.'
+  }
+  if (!tokenName.trim()) {
+    return 'Please add a token name.'
+  }
+  if (!tokenSymbol) {
+    return 'Please add a token symbol.'
+  }
+  return null
+}
+
+function TokensSettingsBYOT() {
+  const fieldsLayout = useFieldsLayout()
+
+  const {
+    config,
+    onBack,
+    onConfigChange,
+    onNext,
+    steps,
+    step,
+  } = useOnboardingState()
+
+  const [formError, setFormError] = useState(null)
+  const [tokenAddress, setTokenAddress] = useState(config.tokens.address)
+  const [tokenData, loadingTokenData, tokenDataError] = useTokenData(
+    isAddress(tokenAddress) && tokenAddress
+  )
+  const [gTokenName, setGTokenName] = useState(config.tokens.name)
+  const [gTokenSymbol, setGTokenSymbol] = useState(config.tokens.symbol)
+
+  const handleTokenAddressChange = useCallback(event => {
+    setFormError(null)
+    setTokenAddress(event.target.value)
+  }, [])
+  const handleTokenNameChange = useCallback(event => {
+    setFormError(null)
+    setGTokenName(event.target.value)
+  }, [])
+  const handleTokenSymbolChange = useCallback(event => {
+    setFormError(null)
+    setGTokenSymbol(event.target.value.trim().toUpperCase())
+  }, [])
+
+  const handleNext = useCallback(
+    event => {
+      event.preventDefault()
+
+      const error = validationError(tokenAddress, gTokenName, gTokenSymbol)
+      setFormError(error)
+
+      if (!error) {
+        onConfigChange('tokens', {
+          address: tokenAddress,
+          name: gTokenName,
+          symbol: gTokenSymbol,
+        })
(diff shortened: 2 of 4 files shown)


src/components/Onboarding/Screens/Apps/TokensSettings.js
```diff
@@ -1,21 +1,15 @@
...
-      {title}
-      <Navigation
-        backEnabled
-        nextEnabled
-        nextLabel={`Next: ${steps[step + 1].title}`}
-        onBack={onBack}
-        onNext={onNext}
-      />
-    </div>
+    <>
```
pavloMandryk, line 9: Is this the way you want the screens to switch? I decided to split the screen into two different because it was getting way too long and the shared code was not that much.

src/components/Onboarding/Screens/Apps/TokensSettings.js
```diff
@@ -1,21 +1,15 @@
...
+      {config.garden.type === 0 && <TokensSettingsNative />}
+      {config.garden.type === 1 && <TokensSettingsBYOT />}
+    </>
```
pavloMandryk, line 12: Is this the way you want the screens to switch? I decided to split the screen into two different because it was getting way too long and the shared code was not that much.

src/hooks/useToken.js
```diff
@@ -23,24 +23,32 @@ export function useTokenBalanceOf(tokenAddress, account) {
...
   const tokenContract = useContractReadOnly(tokenAddress, tokenAbi)
 
   useEffect(() => {
     const fetchTokenData = async () => {
+      if (tokenAddress) {
+        setLoading(true)
+      }
+      const name = await tokenContract.name()
       const decimals = await tokenContract.decimals()
       const symbol = await tokenContract.symbol()
```
pavloMandryk, line 38: If an address is not a token address, with a name and symbol, it crashes. Need some guidance to treat that exception.

pavloMandryk: With the new pushes this bug is solved. As long as I am concerned everything works correctly.

src/components/Onboarding/Screens/Apps/TokensSettingsNative.js
```diff
@@ -0,0 +1,450 @@
...
+  IconPlus,
+  IconTrash,
+  Info,
+  isAddress,
+  RADIUS,
+  TextInput,
+  useTheme,
+} from '@1hive/1hive-ui'
+import Navigation from '../../Navigation'
+import { useOnboardingState } from '../../../../providers/Onboarding'
```
fabriziovigevani, line 18:
```suggestion
import { useOnboardingState } from '@providers/Onboarding'
```

src/components/Onboarding/Screens/Apps/TokensSettingsNative.js
```diff
@@ -0,0 +1,450 @@
...
+  const { onBack, onNext, steps, step, onConfigChange } = useOnboardingState()
+
+  const [formError, setFormError] = useState()
+
+  const [tokenName, setTokenName] = useState('')
+  const [tokenSymbol, setTokenSymbol] = useState('')
+  const [commonPool, setCommonPool] = useState(0)
+  const [accountStake] = useState(0)
+
+  const [members, setMembers] = useState([['', accountStake]])
```
fabriziovigevani, line 69: Same as in Token settings BYOT, we should use the config values.

Also, i don't think the `accountStake` variable is being updated si i think we can remove it and use directly 0 ?

Note that for `holders` we need to update the default value in `providers/Onboarding` to `[['', 0]]` 

```suggestion
  const [tokenName, setTokenName] = useState(config.tokens.name)
  const [tokenSymbol, setTokenSymbol] = useState(config.tokens.symbol)
  const [commonPool, setCommonPool] = useState(config.tokens.commonPool)

  const [members, setMembers] = useState(holders)
```

src/components/Onboarding/Screens/Apps/TokensSettingsNative.js
```diff
@@ -0,0 +1,450 @@
...
+      const error = validationError(tokenName, tokenSymbol, members, commonPool)
+      setFormError(error)
+
+      if (!error) {
+        onConfigChange('tokens', {
+          name: tokenName,
+          symbol: tokenSymbol,
+          holders: members.filter(
+            ([account, stake]) => isAddress(account) && stake > 0
+          ),
```
fabriziovigevani, line 156: Is the filtering necessary? Since we already ensure in `validationError` that all members follow the conditions?

fabriziovigevani (commented):
Nice job @pavloMandryk 

Left some corrections and suggestions. Let me know wdyt!

src/components/Onboarding/Screens/Apps/TokensSettingsBYOT.js
```diff
@@ -0,0 +1,269 @@
...
+  LoadingRing,
+  TextInput,
+  theme,
+} from '@1hive/1hive-ui'
+import Navigation from '../../Navigation'
+import { useOnboardingState } from '@providers/Onboarding'
+import { useTokenData } from '@hooks/useToken'
+import Header from '../../kit/Header'
+
+function useFieldsLayout() {
```
0xGabi, line 19: What you think of using the same spacing for the fields that we have on the Native screen?

Screen Shot 2021-07-06 at 7 21 41 PM

0xGabi (commented):
Looking good, we are getting close! 🙌

I like the BYOT screen. Left a couple of suggestions.

src/components/Onboarding/Screens/Apps/TokensSettingsBYOT.js
0xGabi: vs BYOT
Screen Shot 2021-07-06 at 7 24 45 PM

pavloMandryk: Screenshot from 2021-07-08 00-06-03

Screenshot from 2021-07-08 00-04-05

Made both forms the same width and added a infobox for context.

Maybe should center the loader?

src/components/Onboarding/Screens/Apps/TokensSettingsBYOT.js
```diff
@@ -0,0 +1,273 @@
...
+                placeholder="Token Address"
+                value={tokenAddress}
+                wide
+              />
+            )}
+          </Field>
+          <div
+            css={`
+              display: flex;
+              margin: auto;
```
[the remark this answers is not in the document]
pavloMandryk: As I had to change form width i used a larger margin (2 * GU)
Screenshot from 2021-07-13 20-10-08

src/components/Onboarding/Screens/Apps/TokensSettingsNative.js
pavloMandryk: I deleted it

src/components/Onboarding/Screens/Apps/TokensSettings.js
0xGabi: Yes, well done!

src/components/Onboarding/Screens/Apps/TokensSettingsBYOT.js
0xGabi: Looking great!

0xGabi (approved):
Awesome work @pavloMandryk. I appreciate you kept receptive to our feedback and that you have addressed our suggestions super fast. Despite we drop the ball and took longer for us to fully review your work. It is great to have you on the team 🙌

src/components/Onboarding/Screens/Apps/TokensSettingsNative.js
```diff
@@ -0,0 +1,417 @@
...
+  const updateMember = useCallback((index, updatedAccount, updatedStake) => {
+    setFormError(null)
+    setMembers(members =>
+      members.map((member, i) =>
+        i === index ? [updatedAccount, updatedStake] : member
+      )
+    )
+  }, [])
+
+  const handleNext = useCallback(
```
0xGabi, line 145: One thing that I notice with a final review is that when we do `onBack` we are not storing the config changes if they were any. I don't think this is particularly bad behavior. Just wanted to bring it up to have your oppinions. 

cc @fabriziovigevani

0xGabi: Regardless I think this PR is adding a lot of value. We can create a new one with that specific change on every `onBack` across screens if we decided to.

pavloMandryk: That's right. I think we'll need to create a PR to fix it on every `onBack`. It is not such a rare practice, when using a installation wizard, moving backwards and performing changes. It would be nice to have support for it.

PR #518, opened 2021-06-30
Garden forum url: Search for links

fabriziovigevani:
If the garden forum link is nor present at the root level (`metadata.forum`) then look under links.
This can happen if a garden doesn't necessarily want to have the forum link at the header.


src/utils/garden-utils.js
```diff
@@ -1,7 +1,26 @@
 import daoList from '@1hive/gardens-dao-list'
 import { addressesEqual } from './web3-utils'
 
+const dafaultForumURL = 'https://forum.1hive.org/'
```
rperez89, line 4: Let's use all capital letters as we use for constants

rperez89 (commented):
LGTM! just one small comment

PR #519, opened 2021-07-04
Gardens Metadata: GitHub Api Integration

rperez89:
Couple of comments:

Created this repo test that we can use for you guys to test how this works and checkout the commits, formats etc
https://github.com/1Hive/dao-list-test
The idea that I have is after you guys validate this process is to create the real repo that or modify the existing one, and create a 1hive-gardens GitHub user that is going to be the one that will  be committing to master from the frontend and that user Github token is going to be set on the `REACT_APP_GITHUB_API_TOKEN` env variable.

The `gardens-dao-list` repo is going to be configured in a way where the only user allowed to commit directly to master is `1hive-gardens`, this way we work in an optimistic way and users that creates gardens have they metadata available as soon as they launch it, if the community see that some of that metadata break our covenant a PR to `gardens-dao-list` would be needed to remove the dao from the list.

Note that I left a `Test` button on the screen so you can execute the process since this github publish is going to be done right after the dao is deployed and we get the dao address back since is needed to be included on the dao data.

Also the real `gardens-dao-list` will include the test that we had in the old one so when a dao is removed those test can be run to check all is ok

Each dao entry will generate one commit per asset and one commit into the network file to include the data, the commits will include the Dao name on the commit message so we can have an easy file history to read


0xGabi (approved):

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,566 @@
...
 import Navigation from '../Navigation'
+import { publishNewDao } from '../../../services/github'
 
-function GardenMetadata({ title }) {
-  const { onBack, onNext, step, steps } = useOnboardingState()
+const COMMUNITY_LINK_TYPE = 'community'
+const DOCUMENTATION_LINK_TYPE = 'documentation'
+
+const DAO_LOGO = 'logo_type'
+const DAO_LOGO_MOBILE = 'logo'
```
fabriziovigevani, line 23: I wouldn't call this logo_mobile since it is present in desktop screens as well. 

Let's maybe stick with what we already use `logo` and `logo_type`

If you agree with this, let's also update the function namings
```suggestion
const DAO_LOGO_TYPE = 'logo_type'
const DAO_LOGO = 'logo'
```

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,566 @@
...
+          <p
+            css={`
+              ${textStyle('body1')};
+              color: ${theme.contentSecondary};
+            `}
+          >
+            Fill with your garden information
+          </p>
+        </div>
+      </div>
```
fabriziovigevani, line 213: Not sure if you have merged with gardens already but we can use the Header form the kit that Paulo added

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,566 @@
...
+            width: 100%;
+            marging-bottom: ${2 * GU}px;
+          `}
+        >
+          <TextInput
+            css="width: 100%;"
+            maxLength="15"
+            onChange={handleGardenNameChange}
+            value={formData.name}
+            required
```
fabriziovigevani, line 234:
```suggestion
            maxLength="15"
            onChange={handleGardenNameChange}
            value={formData.name}
            required
            css="width: 100%;"
```

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,566 @@
...
+            marging-bottom: ${2 * GU}px;
+          `}
+        >
+          <TextInput
+            multiline
+            maxLength="120"
+            css="width: 100%;"
+            onChange={handleGardenDescriptionChange}
+            value={formData.description}
+            required
```
fabriziovigevani, line 250:
```suggestion
            multiline
            maxLength="120"
            onChange={handleGardenDescriptionChange}
            value={formData.description}
            required
            css="width: 100%;"
```

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,566 @@
...
+            `}
+          >
+            <div
+              css={`
+                display: flex;
+                align-items: center;
+                justify-content: space-around;
+              `}
+            >
+              <Field label="Dao Logo">
```
fabriziovigevani, line 271: Was thinking if we should update to garden everywhere in the UI?
I think this is the first time we display Dao nomenclature 
```suggestion
              <Field label="Garden Logo">
```

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,566 @@
...
+                  ${textStyle('body4')};
+                  font-weight: 600;
+                `}
+              >
+                Forum
+              </p>
+              <Help>
+                Add a link to your discussion platform - the best way to do this
+                is to create a discourse (add link to discourse site). If you
+                don't,the 1Hive forum will be assigned (add forum link) by
```
fabriziovigevani, line 334:
```suggestion
                don't, the 1Hive forum will be assigned (add forum link) by
```

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,566 @@
...
+          }
+          css={`
+            width: 100%;
+            marging-bottom: ${2 * GU}px;
+          `}
+        >
+          <TextInput
+            css="width: 100%;"
+            onChange={handleForumChange}
+            value={formData.forum}
```
fabriziovigevani, line 347:
```suggestion
            onChange={handleForumChange}
            value={formData.forum}
            css="width: 100%;"
```

src/services/github.js
```diff
@@ -0,0 +1,256 @@
+import { getNetworkName } from '@utils/web3-utils'
+import env from '@/environment'
+
+const NETWORK = getNetworkName().toLowerCase()
+const GITHUB_API_TOKEN = env('GITHUB_API_TOKEN')
+
+export const fetchLatestCommitSha = async () => {
+  const endpoint =
```
fabriziovigevani, line 8: Wondering if it makes sense to have all endpoints in constants above and have a `ENDPOINT_BASE=https://api.github.com/repos/1Hive/dao-list-test` and derive all endpoints from this one?

src/providers/Onboarding.js
```diff
@@ -10,11 +10,12 @@ const DEFAULT_CONFIG = {
     name: '',
     description: '',
     logo: null,
-    logotype: null,
+    logo_type: null,
+    token_logo: null,
```
fabriziovigevani, line 14: Are you implying we use  `token_logo` as well in the `dao-list` repo ?

Note that we are currently using `token: { logo }`

fabriziovigevani (commented):
Really awesome work Ro! I think the process makes sense, really well thought out!

Left some specific comments. Let me know what do you think.

As general corrections:

- We need to make the images available when coming back to this screen. As it is currently, images are not displayed on that scenario.
- Logo types are usually rectangular since  they involve logo and letters, so we shouldn't have the container circled for that case.
- Noticed that when clicking inside the asset container but outside one of the logo containers, the window to select a file pops up. This might be due to some styling issue.
- Noticed that we are adding the garden type value to the `dao-list` which i don't think we want to do that as we get that form the subgraph.
- We are missing all from validations. (name not empty, description not empty, valid urls, etc)

src/providers/Onboarding.js
rperez89: yes the intention is to just have it on the root since we don't have any more attr

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,593 @@
...
+                  <ImageUploader
+                    id={3}
+                    onImageLoaded={handleOnTokenLogoLoaded}
+                    onImageRemoved={handleOnTokenLogoRemoved}
+                  />
+                </Box>
+              </Field>
+            </div>
+          </Box>
+        </Field>
```
0xGabi, line 332:
```suggestion
		<Box
          heading="Assets"
          css={`
            width: 100%;
          `}
        >
          <div
            css={`
              display: flex;
              align-items: center;
              justify-content: space-around;
            `}
          >
            <Field label="Garden Logo Type">
              <Box
                css={`
                  text-align: center;
                  max-height: 143px;
                `}
              >
                <ImageUploader
                  id={1}
                  onImageLoaded={handleOnGardenLogoTypeLoaded}
                  onImageRemoved={handleOnGardenLogoTypeRemoved}
                />
              </Box>
            </Field>
            <Field label="Garden Logo">
              <Box
                css={`
                  text-align: center;
                `}
              >
                <ImageUploader
                  id={2}
                  onImageLoaded={handleOnGardenLogoLoaded}
                  onImageRemoved={handleOnGardenLogoRemoved}
                />
              </Box>
            </Field>
            <Field label="Token Icon">
              <Box
                css={`
                  text-align: center;
                `}
              >
                <ImageUploader
                  id={3}
                  onImageLoaded={handleOnTokenLogoLoaded}
                  onImageRemoved={handleOnTokenLogoRemoved}
                />
              </Box>
            </Field>
          </div>
        </Box>
 ```

Comment (0xGabi) [reply src/components/Onboarding/Screens/GardenMetadata.js:332]: I test and removing the `Field` element prevents the issue that shows the upload modal when clicking any place of the `Field`. Note we will need some padding fixes afterwards.

Comment (0xGabi) [reply src/components/Onboarding/Screens/GardenMetadata.js:332]: While we are at this. What you think of reducing the default size of the `ImageUploader` component now that we are just showing a green check?

File: src/components/Onboarding/Screens/GardenMetadata.js [hunk 11]
```diff
@@ -1,22 +1,593 @@
...
+
+const GARDEN_LOGO_TYPE = 'logo_type'
+const GARDEN_LOGO = 'logo'
+const TOKEN_LOGO = 'token_logo'
+
+const URL_REGEX = new RegExp(
+  /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)/
+)
+
+function GardenMetadata() {
```
0xGabi, line 32: Note we are not indicating to the user that we have store the image data when we go back and forth between screens.

What you think of having a `useEffect` that evaluates the config data we have stored on each type of asset. If it's not the default it means there were assets previously store and we can safely show a green check.

We can evaluate the `setFormData` and check the fields `[type]: base64` for every asset, comparing with the empty value.

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,593 @@
...
+    },
+    [handleOnAssetAdded]
+  )
+
+  const handleOnTokenLogoLoaded = useCallback(
+    (base64, fileExtension) => {
+      handleOnAssetAdded(TOKEN_LOGO, base64, fileExtension)
+    },
+    [handleOnAssetAdded]
+  )
```
0xGabi, line 165: Wonder if you consider using `handleOnAssetAdded` directly instead of creating several partially applied functions.

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,593 @@
...
+
+      {displayErrors && errors.length > 0 && (
+        <Info mode="warning">
+          {errors.map((err, index) => (
+            <div key={index}>{err}</div>
+          ))}
+        </Info>
+      )}
+
+      <Button onClick={() => publishNewDao(formData)}> TEST </Button>
```
0xGabi, line 395: Should we remove it now?

src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -1,22 +1,593 @@
...
+          linksType={DOCUMENTATION_LINK_TYPE}
+          links={formData.links.documentation}
+          onAddLink={addLink}
+          onRemoveLink={removeLink}
+          onUpdateLink={updateLink}
+        />
+      </div>
+
+      {displayErrors && errors.length > 0 && (
+        <Info mode="warning">
```
0xGabi, line 388:
```suggestion
        <Info mode="error">

Comment (0xGabi) [reply src/components/Onboarding/Screens/GardenMetadata.js:388]: To keep consistent with the rest of the screen we should use `error` mode here.

**Review by 0xGabi — CHANGES_REQUESTED** [review]
Well done Ro! 🙌

I left a few final comments and suggestions to add before merging.

File: src/components/Onboarding/Screens/GardenMetadata.js (hunk 10 above)
Comment (rperez89) [reply src/components/Onboarding/Screens/GardenMetadata.js:332]: > I tested and removing the `Field` element prevents the issue that shows the upload modal when clicking any place of the `Field`. Note we will need some padding fixes afterwards.

hey gabi could you double check? just tested and still have the issue, about the size which size do you recommend?

Comment (rperez89) [reply src/components/Onboarding/Screens/GardenMetadata.js:332]: > I tested and removing the `Field` element prevents the issue that shows the upload modal when clicking any place of the `Field`. Note we will need some padding fixes afterwards.

hey gabi, could you double check? just tested what you suggested and still have the issue, about the size, which size do you recommend?

File: src/components/Onboarding/Screens/GardenMetadata.js (hunk 12 above)
Comment (rperez89) [reply src/components/Onboarding/Screens/GardenMetadata.js:165]: this is to avoid recreating the function by using `onClick={() => handleOnAssetAdded(xxx,xxx,xx)}`

File: src/components/Onboarding/Screens/GardenMetadata.js (hunk 13 above)
Comment (rperez89) [reply src/components/Onboarding/Screens/GardenMetadata.js:395]: yes! just left for you to test it

File: src/components/Onboarding/Screens/GardenMetadata.js [hunk 15]
```diff
@@ -385,19 +388,18 @@ function GardenMetadata() {
       </div>
 
       {displayErrors && errors.length > 0 && (
-        <Info mode="warning">
+        <Info mode="error">
```
0xGabi, line 391: Nitpick, there is no padding with the navigation.

0xGabi (approved):
Looking really good now!

PR #522, opened 2021-07-10
Multichain support

rperez89:
@0xGabi created this pr for start a discussion around the best implementation for the multi chain support, from this pr we can replicate stuffs on the Aragon client.

Things to take into consideration:
- created an auto connect bit of code when MetaMask is detected and there is some address authorized for the site to improve the UX
- Since we need to detect the chainId to be able to pull the data from the respective chain I think we need to start asking the user to connect to a wallet to use the app (so we can fetch the garden list from the correct chain), if we want to improve ux we can do something similar to what was done to autoconect, (we can first check if we have MetaMask on windows.ethereum and if is present we can get the chain from there without asking to connect the wallet)
- We need the previous mentioned screen asking to connect to a wallet before getting to the garden list main page or some kind of div with a button asking to connect or something.

With this mentioned points we can replicate on Aragon and get the node based on the `wallet.chainId` property and refetch the data by listening on the providers to the chainId change

We need to have a some  new env vars deployed:

REACT_APP_RINKEBY_ETH_NODE (probably we are not going to support rinkeby on the production deployment but just for now to test)
REACT_APP_XDAI_ETH_NODE 
GITHUB_API_TOKEN (going to create a 1hive-dev account on github, is the one that is going to push changes to the dao-list  repo, for now you can set your own github token to test)


src/providers/Wallet.js
```diff
@@ -14,6 +14,7 @@ function useWalletAugmented() {
 // Adds Ethers.js to the useWallet() object
 function WalletAugmented({ children }) {
   const wallet = useWallet()
+  console.log('Chain Id ', wallet)
```
0xGabi, line 17: Let's remember to remove this.

0xGabi (commented):
I think this is a step in the right direction. We should refactor the code to consider a multichain site. Things like:
- Notify which network the user is connected. We can follow something similar to what Honeycomb UI does:

Screen Shot 2021-07-14 at 4 40 54 PM

- Refactor the `AccountModule` to support multiple chains.
- When there is not wallet provider connected we should:
	- Block the `Grow Garden` button.
	- We can display a message or image that asks the user to "Connect Wallet" in the same place we show the list of Gardens.

src/services/github.js
```diff
@@ -45,14 +44,15 @@ export const fetchBaseTreeSha = async commitSha => {
   }
 }
 
-export const createTree = async (baseTreSha, fileContent) => {
+export const createTree = async (baseTreSha, chainId, fileContent) => {
```
0xGabi, line 47: The GitHub service looks awesome. I notice we have an inconsistent format between the two versions of `dao-list` with regards to how we handle `token_logo` assets.

src/providers/Wallet.js
```diff
@@ -13,23 +20,122 @@ function useWalletAugmented() {
 
 // Adds Ethers.js to the useWallet() object
 function WalletAugmented({ children }) {
+  // This new state is neede to avoid showing the user the profile signature request on every network change
+  const [chainReseted, setChainReseted] = useState(false)
+  const [chainId, setChainId] = useState(-1)
```
0xGabi, line 25: I believe this `chainId` state variable is not relevant anymore with `preferredNetwork`. In places, we would like a reference to the chainId we just use `wallet.chainId`

src/providers/Wallet.js
```diff
@@ -13,23 +20,122 @@ function useWalletAugmented() {
...
+    }
+  }, [chainId, isSupportedChain, wallet, resetConnection])//eslint-disable-line
 
-  const contextValue = useMemo(() => ({ ...wallet, ethers }), [wallet, ethers])
+
+  useEffect(() => {
+    if (isSupportedChain(chainId)) {
+      setPreferredChain(chainId)
+    }
+  }, [chainId])
```
0xGabi, line 111: Why we do not use `wallet.chainId` instead?

src/providers/Wallet.js
```diff
@@ -13,23 +20,122 @@ function useWalletAugmented() {
...
+      if (isSupportedChain(wallet._web3ReactContext.chainId)) {
+        connectNetwork()
+      } else {
+        resetConnection()
+      }
+    }
+    if (isSupportedChain(wallet._web3ReactContext.chainId)) {
+      setChainId(wallet.chainId)
+      setPreferredNetwork(wallet.chainId)
+    }
```
0xGabi, line 103: Would you mind give me a bit of context on what we are doing here? I am not fully understanding what's happening.

src/providers/Wallet.js
```diff
@@ -13,23 +20,122 @@ function useWalletAugmented() {
...
+  }, [chainId, isSupportedChain, wallet, resetConnection])//eslint-disable-line
```
0xGabi, line 104: I think we can remove `chainId` here. Related to the last few comments.

src/providers/Wallet.js
```diff
@@ -13,23 +20,122 @@ function useWalletAugmented() {
 
 // Adds Ethers.js to the useWallet() object
 function WalletAugmented({ children }) {
+  // This new state is neede to avoid showing the user the profile signature request on every network change
+  const [chainReseted, setChainReseted] = useState(false)
```
0xGabi, line 24: I wonder if this is the proper way to handle this side effect. Let's keep it this way, and keep it in our minds once we migrate to IDX.

0xGabi (changes requested):
Good work Ro, it was a big change that affected several places on the codebase and you kept every place updated 💪

I would have appreciated a split on several PRs. Though I recognize might have been difficult to separate the changes.

The flow of the app feels much smoother, and the integration with GitHub to fetch the metadata is working great! 🙌

A place where we can do a better job is with the error screen when a user switches network while exploring a Garden. Maybe we should give the option to continue exploring and disconnect their account. Or return home connected with the new network.

src/assets/providers.svg
```diff
@@ -0,0 +1 @@
...Kou76tFTU9xViZ7K4m4bPapOctiYorJVqIUz5V40KQUsGgIKqhcuTsErVUgNiZa3VFTcVYbesFumdFs2eq3TUSl+lYde++yoxPuFoh5XLj2cYgIyIQD5VFw4occm2j2OKGcSAt4S081Qb1mWPIVQMXqNbtnulolevSTsykOvVip2FD1IBEdI1REo4AKERZ10usH3aC7J+7EHFrB0CQTtE3LZ1NYFu3LQO19X4df4ynxJpRr01EsqiVGPrvVCOSGJaBDSnJEqJlyYJAulWUAEwnIw340sZ/EprgCu1pZeNlA6uXAxqMovp4TJhRunfL/6998fdsgWt7g0w0JLWPNBUWnBREIAhMWdgM+Lk6UpdOLcF2cd/KqVzMT5FFVhb9itNLZ3VGWcrRKzy7h5PVVhr116jhkdhyjsAV7uASaOoGm5cEUYhLweDMpBX1hbTZ/JC8VP9sjdUqccczvsDaVT1xtU75dzJ305S2vuVO9X/176cpb2Sb96vxa3S1S5+YA9jgDcUOIMVz1HJg5AOOPHeYVMYYEBdQtLazu+kEnd+bEu/8mMb2RSdz0eaeLXROavb27PJkea+NVcUdUkMhu31aaPzAIyT0CruYjGSdSW5ar5ytLeXh1fSIl7w8tjrbwdjW+kxL3WnTbUYTuazKTEvfa9NtSRcdhcLSELW2zBGHCjTw4JM3k8EgJIZxXIruWtCdSxnLO3X7332NAhw4zJOed7ypxXOmSYMTnnyZ71Xk2HDDMm55yyeo8+3+rTrywKv6UIHxS+jwVCvoraA66te4YZU+OuLwvC1+gda/xfTMe5Kwhffa5XqNvSWvr3taLQ6RXqtsfhrb31iAGMZqBM98Rtn6/xDKAzKNRtRftB9zwXfY2L7toAZ0fO6XUuqbN+M3ZG+vt11D+b5aKvNpv0jwwYh4vpyoZipAO++IAQ/2oVtlLTXjXXpkIn0tfbjV/jotcdGOXryBnPd+NXv5kbwZxI3+RkN3612YkRzIn0NQP80GPl9MvDINx+MAgjN20ujGcuEvIH627v8vy80eAQBs3zi8te93hwaK5f7xzndH53fX1V5xDW6/Xrm7v5eOy8M9evg37/7OR+NqvVOIRBcza7P5lM+gcGj8PFejq9Xa1s26bzejBorla302ZzUd44/BOhpWfg6dT4kQAAAABJRU5ErkJggg=="/></defs></svg>
```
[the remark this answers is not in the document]
rperez89: Interesting that's true, what do you think we should do?

src/providers/Gardens.js
```diff
@@ -58,31 +58,60 @@ export function useGardens() {
...
+
+  return [
+    useMemo(
+      () =>
+        gardens.map(garden =>
+          mergeGardenMetadata(garden, gardensMetadata, networkId)
+        ),
+      [gardens, gardensMetadata, networkId]
+    ),
+    loading,
```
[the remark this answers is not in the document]
rperez89: oh really  nice catch totally missed this loading

src/assets/providers.svg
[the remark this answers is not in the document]
0xGabi: I think is fine for now. We can ask Fiore to create a new asset more align with the Gardens branding.

src/providers/Wallet.js
```diff
@@ -13,23 +20,122 @@ function useWalletAugmented() {
...
+      await connect()
+    }
+    if (window.ethereum && wallet.chainId === -1) {
+      connectNetwork()
+    }
+    if (isSupportedNetwork) {
+      setChainId(wallet.chainId)
+      setPreferredNetwork(wallet.chainId)
+    }
+  }, []) //eslint-disable-line
```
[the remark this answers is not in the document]
rperez89: that is because when you don't pass any dependency that useEffect is only triggered at the first render time

src/providers/Wallet.js
rperez89: good catch gabi, that state stayed old since when i started to do the first iteration and forgot to delete it

rperez89:
@0xGabi thank you so much for the review made me realize of some old stuffs that i left on the pr when i started  with the wip that i forgot to change, note that will need this <https://github.com/1Hive/dao-list/pull/38> to be merged in order to be tested again will aprecciate so more hard test after this

PR #524, opened 2021-07-20
Onboarding: Add Issuance settings screen

PJColombo:
This PR implements the issuance settings screen. 

It also includes the issuance chart but we still need to generate the data for it. @sembrestels  was working on the [model](https://colab.research.google.com/drive/1vpiqGPgUEa7wk3ibIkyFGnoP3rfIp3Nb#scrollTo=O8EdvSbNZz5z). However, the calculations are not entirely correct and they still need to be validated.


src/components/Onboarding/Screens/Apps/IssuanceSettings/IssuanceChart.js
```diff
@@ -0,0 +1,81 @@
...
+
+const calculateRatio = (
+  initialRatio,
+  targetRatio = 0.001,
+  maxAdjustmentRatioPerYear
+) => {
+  let balance = initialRatio
+  let supply = 1
+  const ratio = [balance / supply]
+  for (let i = 0; i < 365; i++) {
```
0xGabi, line 18: A small comment with an explanation of what is happening here would be apprecited. I kind of understand, thought seems a piece of logic worth giving more context about it.

src/components/Onboarding/Screens/Apps/IssuanceSettings/IssuanceChart.js
```diff
@@ -0,0 +1,81 @@
...
+    const adj = ((1.0 - balance / supply / targetRatio) / 365) * supply
```
0xGabi, line 19: What do you think of adding the extra parameters to improve readability and help understand the logic?

src/components/Onboarding/Screens/Apps/IssuanceSettings/IssuanceChart.js
```diff
@@ -0,0 +1,81 @@
...
+    const maxAdj = maxAdjustmentRatioPerYear / 365 / targetRatio
+    const totalAdj = Math.sign(adj) * Math.min(Math.abs(adj), maxAdj)
+    balance += totalAdj
```
0xGabi, line 22: I don't understand why both `balance` and `supply` are the sum of `totalAdj` 🤔

@sembrestels would you mind clarifying this for me 🙏

0xGabi (approved):
Looking good 🙌 I merge with the `garden` branch to reduce the review files. 

Left a few comments related to the calculation formula.

src/utils/kit-utils.js
```diff
@@ -1,3 +1,4 @@
 export const MINUTE_IN_SECONDS = 60
 export const HOUR_IN_SECONDS = MINUTE_IN_SECONDS * 60
 export const DAY_IN_SECONDS = HOUR_IN_SECONDS * 24
+export const YEARS_IN_SECONDS = 365 * DAY_IN_SECONDS
```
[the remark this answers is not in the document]
sembrestels: Good point! The parameter is defined as `maxAdjustmentPerSecond` and we are using this same number in the Issaunce tests in order have this parameter annualized. However, as I posted in [this section of my post on the dynamic issuance v1](https://forum.1hive.org/t/insights-on-dynamic-issuance-after-modeling-it/4187#parameter-name-max-adjustment-per-year-may-be-misleading), this annualization is misleading, so I'm changing the parameter name to **Throttle** in the frontend, and I use another parameter in the Issuance v2: **Recovery Time**.

sembrestels:
@0xGabi I added comments in the code so it's now easier to follow. Thanks for pointing it out!

PR #525, opened 2021-07-25
Side panel 

pavloMandryk:
This PR contains the reorganization of the steps, so we can have the same order as shows on issue[ #217 ](https://github.com/1Hive/gardens/issues/217), and the implementation of substeps.

### DIFF
--- a/src/components/Onboarding/Screens/config.js
+++ b/src/components/Onboarding/Screens/config.js
@@ -10,53 +10,51 @@ import TokensSettings from './Apps/TokensSettings'
 import VotingSettings from './Apps/VotingSettings'
 
 const STEPS = [
-  'Select garden type',
-  'Configure garden',
+  'Select Garden type',
+  'Add Garden metadata',
+  'Configure tokenomics',
+  'Configure governance',
   'Review information',
   'Launch Garden',
 ]
 
 export const Screens = [
   {
-    key: STEPS[0],
+    parent: STEPS[0],
     title: 'Type selection',
     Screen: GardenTypeSelector,
   },
-  { key: STEPS[1], title: 'Garden metadata', Screen: GardenMetadata },
+  { parent: STEPS[1], title: 'Garden metadata', Screen: GardenMetadata },
+  { parent: STEPS[2], title: 'Garden token', Screen: TokensSettings },
   {
-    key: STEPS[1],
+    parent: STEPS[2],
     title: 'Honeyswap Liquidity',
     Screen: HoneyswapLiquidity,
   },
   {
-    key: STEPS[1],
-    title: 'Tokens settings',
-    Screen: TokensSettings,
-  },
-  {
-    key: STEPS[1],
-    title: 'Voting settings',
-    Screen: VotingSettings,
-  },
-  {
-    key: STEPS[1],
-    title: 'Conviction Voting settings',
-    Screen: ConvictionVotingSettings,
-  },
-  {
-    key: STEPS[1],
-    title: 'Issuance settings',
+    parent: STEPS[2],
+    title: 'Issuance policy',
     Screen: IssuanceSettings,
   },
   {
-    key: STEPS[1],
-    title: 'Covenant settings',
+    parent: STEPS[3],
+    title: 'Community Convenant',
     Screen: AgreementSettings,
   },
   {
-    key: STEPS[2],
+    parent: STEPS[3],
+    title: 'Conviction voting',
+    Screen: ConvictionVotingSettings,
+  },
+  {
+    parent: STEPS[3],
+    title: 'Decision voting',
+    Screen: VotingSettings,
+  },
+  {
+    parent: STEPS[4],
     title: 'Review information',
     Screen: ReviewInformation,
   },
-  { key: STEPS[3], title: 'Launch Garden', Screen: LaunchGarden },
+  { parent: STEPS[5], title: 'Launch Garden', Screen: LaunchGarden },
 ]
--- a/src/components/Onboarding/Steps/Step.js
+++ b/src/components/Onboarding/Steps/Step.js
@@ -1,0 +1,44 @@
+import React, { useMemo } from 'react'
+import StepItem from './StepsItem'
+import { useOnboardingState } from '@providers/Onboarding'
+
+function Step({ currentStep, label, step, stepNumber, substeps }) {
+  const { step: onboardingStep } = useOnboardingState()
+  const [displayedSubsteps] = useMemo(() => {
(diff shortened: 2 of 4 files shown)


src/components/Onboarding/Screens/config.js
```diff
@@ -23,40 +25,36 @@ export const Screens = [
     Screen: GardenTypeSelector,
   },
   { key: STEPS[1], title: 'Garden metadata', Screen: GardenMetadata },
+  { key: STEPS[2], title: 'Tokens settings', Screen: TokensSettings },
```
rperez89, line 28: i think that makes more sense to rename this `key` to `parent` what do you think?

src/components/Onboarding/Steps/Step.js
```diff
@@ -0,0 +1,44 @@
...
+import { useOnboardingState } from '@providers/Onboarding'
+
+function Step({ currentStep, label, step, stepNumber, substeps }) {
+  const { step: onboardingStep } = useOnboardingState()
+  const [displayedSubsteps] = useMemo(() => {
+    const displayedSubsteps = substeps.map((step, index) => {
+      if (substeps.length > 0) {
+        const label = step[0].title
+
+        const statusIndex = step[1]
```
rperez89, line 12: Trying to understand the a bit the logic here, based on the naming what should be the statusIndex variable since is the value `index - hiddenCount` right? what status is that indicating?

src/components/Onboarding/Steps/StepsItem.js
```diff
@@ -31,17 +31,18 @@ function ConfigureStepsItem({ stepNumber, step, label, currentStep }) {
       css={`
         display: flex;
         align-items: center;
-        height: ${5 * GU}px;
+        height: ${type === 'step' ? 5 * GU : 2 * GU}px;
```
rperez89, line 34: since we only have two kinds and we are re writing the `type === 'step'` check, we can have a variable outside called `const isStepType = type === 'step'` and then just do  `height: ${isStepType? 5 * GU : 2 * GU}px;`

rperez89 (commented):
Awesome man! just some small comments

src/components/Onboarding/Screens/config.js
```diff
@@ -23,40 +25,36 @@ export const Screens = [
...
+    key: STEPS[2],
+    title: 'Issuance settings',
+    Screen: IssuanceSettings,
   },
   {
-    key: STEPS[1],
-    title: 'Voting settings',
-    Screen: VotingSettings,
+    key: STEPS[3],
+    title: 'Covenant settings',
```
0xGabi, line 41:
```suggestion
    title: 'Community Covenant',

File: src/components/Onboarding/Screens/config.js [hunk 5]
```diff
@@ -23,40 +25,36 @@ export const Screens = [
...
+    Screen: AgreementSettings,
   },
   {
-    key: STEPS[1],
+    key: STEPS[3],
     title: 'Conviction Voting settings',
```
0xGabi, line 46:
```suggestion
    title: 'Conviction voting',

File: src/components/Onboarding/Screens/config.js [hunk 6]
```diff
@@ -23,40 +25,36 @@ export const Screens = [
...
-    key: STEPS[1],
-    title: 'Issuance settings',
-    Screen: IssuanceSettings,
-  },
-  {
-    key: STEPS[1],
-    title: 'Covenant settings',
-    Screen: AgreementSettings,
+    key: STEPS[3],
+    title: 'Voting settings',
```
0xGabi, line 51:
```suggestion
    title: 'Decision voting',

File: src/components/Onboarding/Steps/StepsPanel.js [hunk 7]
```diff
@@ -52,10 +64,7 @@ function StepsPanel() {
           height: ${25 * GU}px;
         `}
       >
-        <CircleGraph
-          value={displayedSteps[step][1] / (displayedStepsCount - 1)}
-          size={25 * GU}
-        />
+        <CircleGraph value={step / Screens.length} size={25 * GU} />
```
0xGabi, line 67: I belive we need to use `Screens.lenght - 1` here.

0xGabi (changes requested):
@pavloMandryk great work. Super smooth. I left several cosmetic changes to the name descriptions.

One thing that I am not convinced is about the `CircleGraph` component showing both the `%` of the progress and also the `1/6` step we are. Wonder if we shoudl keep the steps there 🤔

pavloMandryk:
I have made some changes based on your suggestions:
- Changed `key` to `parent` in the Screens array. Found a bit confusing the `key` naming.
- Took off the `1/6` from the `<CircleGraph />`. It wasn't providing useful info as we already have the side menu to show in which step the user currently is. Also it wasn't taking into account substeps, so I think it is a good idea to get rid of it.
- Changed the steps and substeps naming.
- Had to ajust the font size of substeps so the new names could fit better.
- Created a constant isTypeStep to  check the type  of the input to `<StepsItem />`.

src/components/Onboarding/Screens/config.js
GriffGreen: Can we stick with Disputable Voting?

src/components/Onboarding/Steps/Step.js
rperez89: i think we just miss this small comment, just to understand

src/components/Onboarding/Steps/StepsPanel.js
```diff
@@ -52,36 +64,23 @@ function StepsPanel() {
...
-            position: absolute;
-            top: 130px;
-            font-size: 20px;
-            color: #8e97b5;
-            opacity: 0.7;
-          `}
-        >
-          {`${displayedSteps[step][1] + 1}/${displayedStepsCount}`}
-        </div>
+        <CircleGraph value={step / (Screens.length - 1)} size={25 * GU} />
```
0xGabi, line 67:
```suggestion         	
        <CircleGraph value={step / Screens.length} size={25 * GU} />

0xGabi (approved):
Well done! Looking great. Let's merge it.

src/components/Onboarding/Steps/StepsPanel.js
0xGabi: After the new changes, I notice it was displaying wrong. Going back to your first implementation worked fine.

0xGabi: I'll go ahead and commit this change and merge the PR 🙌

PR #534, opened 2021-08-07
App: add progress save toast

0xGabi:
https://user-images.githubusercontent.com/9082013/128612928-922bd173-f91c-459a-9b55-6bfa480287eb.mov

### DIFF
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -1,13 +1,17 @@
+import { useToast } from '@1hive/1hive-ui'
 import React, { useCallback, useState } from 'react'
 import styled from 'styled-components'
+
+import { useNodeHeight } from '@hooks/useNodeHeight'
+
 import GardensList from './GardensList'
 import Onboarding from './Onboarding'
 import LandingBanner from './LandingBanner'
-import { useNodeHeight } from '@hooks/useNodeHeight'
 
 function Home() {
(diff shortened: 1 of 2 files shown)


fioreb:
The functionality is perfect!
But can we make it look link this: https://www.figma.com/file/ySHlXMwz5elSXKI3feisJU/Gardens-wizard-creation-UI?node-id=543%3A13865 ? only with padding and not that long. Thanks 🙏

src/components/Home.js
```diff
@@ -1,21 +1,26 @@
+import { useToast } from '@1hive/1hive-ui'
```
fabriziovigevani, line 1: Let's place this import after `styled-components` import.

We usually have imports from our own libraries after general library imports.

src/components/Home.js
```diff
@@ -1,21 +1,26 @@
...
 import React, { useCallback, useState } from 'react'
 import styled from 'styled-components'
+
+import { useNodeHeight } from '@hooks/useNodeHeight'
```
fabriziovigevani, line 5: What was the rationale for moving this up? 

We usually have local hooks after components.

fabriziovigevani (commented):
🚀 🚀 🙌

fabriziovigevani (approved):
💪

PR #537, opened 2021-08-11
feat: add env var for xDai RPC

crisog:
Using public RPCs or hard-coded private RPCs is not optimal for our hosted frontends. They could easily stop working or be attacked.

This adds an environment variable for the RPC, this will let us use a private RPC without publicly exposing it.

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -24,7 +24,7 @@ const ENV_VARS = {
     return process.env.REACT_APP_ETH_NODE || ''
   },
   XDAI_NODE() {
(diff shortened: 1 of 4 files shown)


0xGabi (approved):

fabriziovigevani (commented):
Ah we should actually use ETH_NODE which is used for this purpose and add the env var on vercel (note that we already do that for Rinkeby)

fabriziovigevani:
@crisog @0xGabi 

Updated the code to use ETH_NODE. Already added vercel en vars to reflect this change

PR #541, opened 2021-08-15
Chore: sort Gardens by proposal count by default

0xGabi:


0xGabi:
Close it as we will sort by liquidity instead.

PR #543, opened 2021-08-18
Refactor(hooks): useMounted logic 

0xGabi:
As explained in: https://kentcdodds.com/blog/useeffect-vs-uselayouteffect#one-special-case we want to change useEffect for useLayoutEffect on this particular case.

### DIFF
--- a/src/hooks/useMounted.js
+++ b/src/hooks/useMounted.js
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useCallback } from 'react'
+import { useLayoutEffect, useRef, useCallback } from 'react'
 
 // Simple hook for checking a component is mounted prior to an async state update
 export function useMounted() {
@@ -6,7 +6,7 @@ export function useMounted() {
 
   const getMounted = useCallback(() => mounted.current, [])
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     return () => {
       mounted.current = false
     }


rperez89:
> As explained in https://kentcdodds.com/blog/useeffect-vs-uselayouteffect#one-special-case we want to change `useEffect` for `useLayoutEffect` in this particular case.

@0xGabi as far as i understood from kent `However, if your effect is mutating the DOM (via a DOM node ref) and the DOM mutation will change the appearance of the DOM node between the time that it is rendered and your effect mutates it`. 

And in this case we are not mutating the DOM since that ref is not being assigned to any DOM element like when we do: 


``` 
 const buttonRef = useRef();
  return(
    <div >
      <h1>The Button has a ref</h1>  
      <button
        onClick={() => {
          console.log(buttonRef.current);
        }}
        ref={buttonRef}
        >
          Special Button
       </button>
    </div> ````

I might be wrong but i think is no needed

0xGabi:
What you mention is right. Thought, did you read the [One special case section](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect#one-special-case)? I believe we are exactly on that case when we are managing a reference across hooks that need to know if the component was mounted.

rperez89:
> What you mention is right. Thought, did you read the [One special case section](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect#one-special-case)? I believe we are exactly on that case here 🤔

wondering if you tried the way it's now and added a console.log to see if the result is an old one? because if we are going to change this one and is not working as expected i think it's need to be changed in all the places where we are updating a ref with an useEffect

fabriziovigevani:
> What you mention is right. Thought, did you read the [One special case section](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect#one-special-case)? I believe we are exactly on that case here 🤔

I think it´s not the same case exactly. In the example it´s updating a ref when the effect runs (which presumably this ref will be used elsewhere and we need that value updated synchronously and before other code runs). Our case it´s different since we are updating the ref on the effect cleanup (when the component unmounts). And this ref is intended to be used within functions which check if the component is unmounted. If component is unmounted, we don´t really care about other code being run afterwards as the last bit of code that will run is the one checking for the `mounted`function.

That said, we should definitely review all places we use refs to the DOM and have an effect 🙌

PR #544, opened 2021-08-19
header  and footer covenant button

rperez89:
Screen Shot 2021-08-19 at 12 15 43 AM

Screen Shot 2021-08-19 at 12 28 05 AM

### DIFF
--- a/src/components/Garden/Footer.js
+++ b/src/components/Garden/Footer.js
@@ -159,7 +159,20 @@ function FixedFooter({ token, gardenAddress }) {
               label="Home"
             />
             <FooterItem
-              icon={<IconWrite alt="covenant" />}
+              icon={
+                <div
+                  css={`
+                    display: flex;
+                    align-items: center;
+                    border-radius: 50px;
+                    width: 24px;
+                    height: 24px;
+                    border: 2px solid ${theme.content};
+                  `}
+                >
+                  <IconWrite alt="covenant" />
+                </div>
+              }
               label="Convenant"
               onClick={handleOnGoToCovenant}
             />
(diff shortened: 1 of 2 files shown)


src/components/Header/Header.js
```diff
@@ -134,20 +134,17 @@ function Header() {
 
 function GardenNavItems({ garden }) {
   const theme = useTheme()
+  const history = useHistory()
   const token = garden.wrappableToken || garden.token
 
+  const handleOnGoToCovenant = useCallback(() => {
```
0xGabi, line 140: What do you think of moving the logic to the hooks directory to DRY?

0xGabi (approved):
LGTM!

src/components/Header/Header.js
rperez89: yeah i thought about it but i think that it doesn't make too much sense to create a hook just redirect to another  page, if you feel ok with that feel free to merge it

src/components/Garden/Footer.js
```diff
@@ -107,14 +112,19 @@ function Footer() {
...
   const theme = useTheme()
+  const history = useHistory()
   const { account } = useWallet()
   const { layoutName } = useLayout()
   const [createProposalModalVisible, setCreateProposalModalVisible] = useState(
     false
   )
 
+  const handleOnGoToCovenant = useCallback(() => {
+    history.push(`/garden/${gardenAddress}/covenant`)
```
fabriziovigevani, line 125: We can use `buildGardenPath` so that it´s not necessary to pass the gardenAddress as prop.

src/components/Garden/Footer.js
```diff
@@ -148,6 +158,24 @@ function FixedFooter({ token }) {
...
+                    border-radius: 50px;
+                    width: 24px;
+                    height: 24px;
+                    border: 2px solid ${theme.content};
+                  `}
+                >
+                  <IconWrite alt="covenant" />
+                </div>
+              }
+              label="Convenant"
```
fabriziovigevani, line 176: `Covenant`

PR #545, opened 2021-08-19
Fix(hooks): useAgreement

0xGabi:
- [x] Fix https://github.com/1Hive/gardens/issues/282

Prevent the App to raise an error with: `Unhandled Rejection (TypeError): Cannot read property 'find' of undefined` because the app object is not defined yet.


fabriziovigevani (approved):
LGTM!

PR #547, opened 2021-08-19
Style: do not allow to create a Garden on mobile

0xGabi:


0xmiel:
Looks good.

I'm wondering whether we should actually remove that button for the time being; at least until we document best practices + get the UX right.

thoughts?

0xGabi:
You are right, I'll add the new changes.

PR #548, opened 2021-08-19
Style(footer): update mobile home button

0xGabi:
Screen Shot 2021-08-19 at 7 58 15 PM


fabriziovigevani (approved):
LGTM

rperez89:
@0xGabi @fabriziovigevani in this case unless that was something brought up by  the ux research i would keep the `Home` naming convention to go back to the main screen

0xGabi:
@rperez89 to give you more context about my decision. I notice the `Home` button redirect to the Garden dashboard view. And that's why I update it in this way. But I agree that having an extra button to go to the Garden `Home` is important too.

PR #557, opened 2021-08-23
 Style: update header background img & modal placeholders

0xGabi:


src/assets/modal-background.svg
```diff
@@ -1 +1,38 @@
...
\ No newline at end of file
+<svg width="726" height="64" viewBox="0 0 726 64" fill="none" xmlns="http://www.w3.org/2000/svg">
```
fabriziovigevani, line 1: We forgot to optimize the asset

0xGabi: Oh good catch, forgot to do that on a few others as well. I'll do it for all of them on a new PR.

PR #558, opened 2021-08-24
Onboarding: review voting screen

0xGabi:
Screen Shot 2021-08-23 at 9 07 19 PM

Screen Shot 2021-08-23 at 9 07 24 PM

### DIFF
--- a/src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
+++ b/src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
@@ -39,7 +39,6 @@ const AdvancedSettingsModal = ({
         }
         duration={voteQuietEndingExtension}
         onUpdate={handleQuiteEndingExtensionPeriodChange}
-        required
       />
       <DurationFields
         label={
@@ -55,7 +54,6 @@ const AdvancedSettingsModal = ({
         }
         duration={voteDelegatedVotingPeriod}
         onUpdate={handleDelegatedVotingPeriodChange}
-        required
       />
     </Modal>
   )
--- a/src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
+++ b/src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
@@ -1,5 +1,6 @@
-import { Help } from '@1hive/1hive-ui'
 import React, { Fragment } from 'react'
+import { Help } from '@1hive/1hive-ui'
+
 import { DurationFields, Modal } from '@components/Onboarding//kit'
 
 const AdvancedSettingsModal = ({
--- a/src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
+++ b/src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
@@ -1,0 +1,64 @@
+import { Help } from '@1hive/1hive-ui'
+import React, { Fragment } from 'react'
+import { DurationFields, Modal } from '@components/Onboarding//kit'
+
+const AdvancedSettingsModal = ({
+  voteQuietEndingExtension,
+  handleQuiteEndingExtensionPeriodChange,
+  voteDelegatedVotingPeriod,
+  handleDelegatedVotingPeriodChange,
+  visible,
+  onClose,
+}) => {
+  const handleDoneClick = () => {
+    onClose()
+  }
+
+  return (
+    <Modal
+      title="Advanced Settings"
+      visible={visible}
+      onClick={handleDoneClick}
+      onClose={onClose}
+      width="100"
+    >
+      <DurationFields
+        label={
+          <Fragment>
+            Quite Ending Extension
+            <Help hint="What is Quite Ending Extension?">
+              <strong>Quite Ending Extension</strong> is the voting duration
+              extension that triggers only from a vote outcome flipping during
+              the Quiet Ending Period. This will add the specified amount of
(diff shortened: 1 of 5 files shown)


src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
```diff
@@ -0,0 +1,64 @@
...
+              time to the vote duration allowing any voters (except stewards)
+              more time to vote. If an outcome is flipped again during the Quiet
+              Ending Extension another extension will trigger. There is no limit
+              on how many times an extension can be triggered.
+            </Help>
+          </Fragment>
+        }
+        duration={voteQuietEndingExtension}
+        onUpdate={handleQuiteEndingExtensionPeriodChange}
+        required
```
rperez89, line 42: I think this required fields are not  needed  to be required since we have default  values on it

src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
```diff
@@ -0,0 +1,64 @@
...
+              <strong>Delegated Voting Period</strong> is the period of time,
+              within the Vote Duration, when stewards can cast votes that have
+              been delegated to them. When this period ends stewards can no
+              longer vote.
+            </Help>
+          </Fragment>
+        }
+        duration={voteDelegatedVotingPeriod}
+        onUpdate={handleDelegatedVotingPeriodChange}
+        required
```
rperez89, line 58: same as above here

src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
```diff
@@ -0,0 +1,64 @@
...
+      />
+    </Modal>
```
rperez89, line 60: wondering if `done editing` is the right wording for the button? maybe just  `Update` or  something like that is better , a two words button seems  a bit  weird

rperez89 (changes requested):

rperez89:
Also there  is a weird bug when etering big amounts, following  some  screenshots

Screen Shot 2021-08-24 at 2 40 01 AM

entering again to the advanced mode : 
Screen Shot 2021-08-24 at 2 40 09 AM

entering again:  

Screen Shot 2021-08-24 at 2 40 22 AM

0xGabi:
@rperez89 I do not detect anything wrong with the behaviour when using big amounts. It's expected when you input an hour value greater than 23 it will be converted to the equivalent number of days.

src/components/Onboarding/Screens/Apps/VotingSettings/AdvancedSettingsModal.js
```diff
@@ -0,0 +1,62 @@
+import { Help } from '@1hive/1hive-ui'
+import React, { Fragment } from 'react'
```
fabriziovigevani, line 2:
```suggestion
import React, { Fragment } from 'react'
import { Help } from '@1hive/1hive-ui'

```

src/components/Onboarding/Screens/Apps/VotingSettings/index.js
```diff
@@ -141,37 +142,42 @@ function VotingSettings() {
...
       setFormError(null)
-      updateField(['voteExecutionDelay', value])
+      updateField(['voteQuietEndingExtension', value])
     },
     [updateField]
   )
 
+  const handleCloseModal = useCallback(() => setOpenSettingsModal(false), [
+    setOpenSettingsModal,
+  ])
```
fabriziovigevani, line 179: There´s no need to add the state updater function, it´s the same function between re-renders.

```suggestion
  const handleCloseModal = useCallback(() => setOpenSettingsModal(false), [])
```

src/components/Onboarding/Screens/Apps/VotingSettings/index.js
```diff
@@ -276,15 +240,70 @@ function VotingSettings() {
...
-          itemHeight={38 * GU}
-          itemSpacing={2 * GU}
-          items={carouselItems}
-          onItemSelected={handleItemSelected}
-        />
-      </div>
+      <DurationFields
+        label={
+          <Fragment>
+            Voting Duration
```
fabriziovigevani, line 246: Should we update in favor of  `Vote Duration`? Since that´s what we are using in the `Help` component. 

Or we could update the hint for `Voting Duration`

src/components/Onboarding/Screens/Apps/VotingSettings/index.js
```diff
@@ -276,15 +240,70 @@ function VotingSettings() {
...
+      <Button
+        size="mini"
+        css={`
+          align-self: flex-end;
+          float: right;
+          margin-bottom: ${5 * GU}px;
+        `}
+        label="Advanced..."
+        onClick={() => setOpenSettingsModal(true)}
+      />
```
fabriziovigevani, line 296: I would avoid using `float` for layout and the `align-self` style is not doing anything in this case. We can wrap the button with a div and apply styles to it.

Also, what do you think of using a `handleOpenModal` callback as we have for closing the modal?



```suggestion 
      <div 
         css={`  
            display: flex;
            align-items: flex-end; 
            margin-bottom: ${5 * GU}px;
          `}
       >
         <Button
           size="mini"
           label="Advanced..."
           onClick={handleOpenModal}
         />
     </div>
```

fabriziovigevani (commented):
Awesome job 🚀 

On top of Rodri´s comments, left some more corrections and suggestions.

PR #560, opened 2021-08-24
Incentivised price oracle

pavloMandryk:
This PR implements the Incentivised Price Oracle functionality, which gives the opportunity to get a reward for updating the Oracle Price on the home page of each garden. It contains:

- ABI update to include the `update` and `canUpdate` functions. [This](https://github.com/1Hive/uniswap-v2-periphery/blob/master/contracts/examples/IncentivisedSlidingWindowOracle.sol) is the contract we are using. 
 Here are the [update()](https://github.com/1Hive/uniswap-v2-periphery/blob/0ca50d6f3f74330d3efc03e6b18ef45d105529f3/contracts/examples/IncentivisedSlidingWindowOracle.sol#L110)  and [canUpdate()](https://github.com/1Hive/uniswap-v2-periphery/blob/0ca50d6f3f74330d3efc03e6b18ef45d105529f3/contracts/examples/IncentivisedSlidingWindowOracle.sol#L95)  functions.

- Update @1hive/connect-gardens version.

- Rename `useRequiereAmount()` hook to `usePriceOracle()` and add calls to update the price oracle and to check if it's updateable, as it's only allowed to update every three hours.

- Update `<PriceToken>` component to create the option to show a `oracleMode` which displays Price Oracle price and provides a button to update it.

It is required to merge the PR [#21](https://github.com/1Hive/1hive-ui/pull/21) in the 1hive-ui repo containing the DotSwitch component used to switch between normal mode (this shows the price from HoneySwap) and oracle mode to make this work.

*At the moment there are two gardens in rinkeby whose incentivisedPriceOracle address we could index in the subgraph:
0xf90cb5fe04ee9cffa034f94285a917143496c8a8 and 0xbb7cc1fd15ee7071f93658d5c46fee41433a5a0b *

**_Update_**

The DotSwitch component from the [1hive-ui PR](https://github.com/1Hive/1hive-ui/pull/21) has been moved to `Metrics.js` file. @rperez89 wdyt?


src/hooks/usePriceOracle.js
```diff
@@ -0,0 +1,81 @@
...
+        if (mounted()) {
+          setConvertedAmount(new BigNumber(result.toString()))
+        }
+      } catch (err) {
+        console.error(`Error consulting converted amount ${err}`)
+      }
+      setLoading(false)
+    }
+
+    const canUpdate = async () => {
```
pavloMandryk, line 50: Should I place this function in its own hook?

fabriziovigevani:
@pavloMandryk will make a full review later today, sorry for the delay!

On a general comment, was wondering about the utility of the component added to the 1hive-ui repo. I think we could just integrate it here for now since i don´t see it as a component that is of general use. 

But want to also know what others think cc. @rperez89 @0xGabi

pavloMandryk:
> @pavloMandryk will make a full review later today, sorry for the delay!
> 
> On a general comment, was wondering about the utility of the component added to the 1hive-ui repo. I think we could just integrate it here for now since i don´t see it as a component that is of general use.
> 
> But want to also know what others think cc. @rperez89 @0xGabi

@fabriziovigevani perfect! I'm having some issues calling the update function on the oracle, I'm trying to fix it. Tell me it runs for you. 
I have discussed the ui component with @0xGabi and we agree on putting it here.

src/components/Garden/Metrics.js
```diff
@@ -78,6 +71,40 @@ const Metrics = React.memo(function Metrics({
   )
 })
 
+function PriceSection({ token, onExecuteIssuance, currency }) {
+  const { layoutName } = useLayout()
+  const compactMode = layoutName === 'small'
+  const [oracleMode, setOracleMode] = useState(false)
+
+  const handleShowOracleMode = () => {
```
fabriziovigevani, line 79: Let´s wrap in `useCallback`

src/components/Garden/Metrics.js
```diff
@@ -128,37 +155,134 @@ function TokenBalance({ label, token, value, currency }) {
...
+
+  const [updatePriceOracle] = useUpdatePriceOracle(token, stableToken)
+
+  useEffect(() => {
+    if (loading) {
+      return
+    }
+
+    setPrice(oracleMode ? convertedAmount.valueOf() : honeyswapPrice)
+  }, [oracleMode, convertedAmount, honeyswapPrice, price, loading])
```
fabriziovigevani, line 182: In this case since we are doing only a calculation, we can memoize the value with a `useMemo` and we can remove the `price` state variable

```
  const price = useMemo(() => {
    if (loading) {
      return 0
    }

    return oracleMode ? convertedAmount.valueOf() : honeyswapPrice
  }, [convertedAmount, honeyswapPrice, loading, oracleMode])
```

src/hooks/usePriceOracle.js
```diff
@@ -0,0 +1,81 @@
+import { useCallback, useEffect, useState } from 'react'
+import { useGardens } from '@providers/Gardens'
+import { useContract, useContractReadOnly } from './useContract'
+import { useMounted } from './useMounted'
```
fabriziovigevani, line 4:
```suggestion
import { useCallback, useEffect, useState } from 'react'
import { useContract, useContractReadOnly } from './useContract'
import { useGardens } from '@providers/Gardens'
import { useMounted } from './useMounted'
```

src/hooks/usePriceOracle.js
```diff
@@ -0,0 +1,81 @@
...
+        console.error(`Error fetching if Price Oracle can update: ${err}`)
+      }
+    }
+    fetchConvertedAmount()
+    fetchCanUpdate()
+  }, [tokenIn, tokenOut, priceOracleContract, amount, mounted, stable])
+  return [convertedAmount, loading, canUpdate]
+}
+
+export function useUpdatePriceOracle(tokenA, tokenB) {
```
fabriziovigevani, line 64: I think we should have this function under `useActions`. 

All functions that perform a contract tx should be placed there. We could also take advantage of the Multimodal component so we get the tx status modal when performing the tx.

fabriziovigevani (commented):
Awesome work!

Left a few comments.

Could you describe the issue you are having when calling update?

0xGabi (approved):
I did a final review. Merging with the latest changes of the `gardens` branch. I notice a couple of details we were missing:
1. When we were using the `usePriceOracle` hook we do not have the HNY price oracle data from the subgraph. So I create a PR to hardcode it: https://github.com/1Hive/gardens/pull/382

2. The order of the tokens when we call the `consult` method depends on the amount we are evaluating. In this case, we want a stable amount of 1 whole token.

3. We were having a strang warning from `react` about using this syntax:
```
       onClick={
         !isActive &&
         (e => {
           e.preventDefault()
           onChange()
         })
```

0xGabi:
I test with both 1Hive and BrighId on xDai and the oracle is updated. For example, this Tenderly transaction: https://dashboard.tenderly.co/tx/xdai/0x6ce8ae798e4d712b5a95154cdac0edb15c33ed8bbd386bd963d8e2829039b9d1

I think this is a key feature that projects like BrightId need while they set up a proper service. So let's go ahead and merge it.

Awesome work @pavloMandryk

PR #566, opened 2021-08-27
UX: Fix withdraw flow

fabriziovigevani:
Uses same transaction feedback modal screen as used for staking and changing support.


0xmiel:
Sweet, this looks good!

PR #567, opened 2021-08-27
UX: Close modal after success

fabriziovigevani:
closes https://github.com/1Hive/gardens/issues/306

Closes modal stepper after all steps have completed.

- Time to pass between step completed and modal closes is set to 2 seconds
- This behavior is omitted if `onCompleteActions !== null`

### DIFF
--- a/src/components/Garden/ModalFlows/SignAgreementScreens/SignAgreementScreens.js
+++ b/src/components/Garden/ModalFlows/SignAgreementScreens/SignAgreementScreens.js
@@ -19,7 +19,7 @@ function SignAgreementScreens({ versionId }) {
     history.push(path)
   }, [history])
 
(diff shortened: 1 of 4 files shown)


0xmiel:
Great! Could we add the same behaviour for when one initially adds support?

Screenshot 2021-08-28 at 12 35 15

Right now we see these buttons, but I think we can remove them and automagically close

fabriziovigevani:
> Great! Could we add the same behaviour for when one initially adds support or challenges a proposal?

Ah yes! Note that this behavior is generally done at the `Stepper` component level meaning that every action performed on the garden that makes use of the modal flow will have this behavior. And this includes the actions you mentioned.

0xGabi (approved):

PR #568, opened 2021-08-27
Homepage stats enhancements + helptip improvements

0xmiel:
- [x] Fix https://github.com/1Hive/gardens/issues/297
- [x] Address https://github.com/1Hive/gardens/issues/275
- [x] Improve helptips

### DIFF
--- a/src/components/Garden/Home.js
+++ b/src/components/Garden/Home.js
@@ -121,7 +121,7 @@ const Home = React.memo(function Home() {
                   margin: ${(below('medium') ? 0 : 3) * GU}px;
                 `}
               >
-                {!compactMode && (
+                {!(layoutName === 'small') && (
                   <Metrics
                     commonPool={commonPool}
                     onExecuteIssuance={actions.issuanceActions.executeIssuance}
--- a/src/components/Garden/Metrics.js
+++ b/src/components/Garden/Metrics.js
@@ -88,15 +88,22 @@ function Metric({ label, value, color, helptip }) {
 
   return (
     <>
-      <p
+      <div
         css={`
           color: ${theme.contentSecondary};
           margin-bottom: ${0.5 * GU}px;
         `}
       >
         {label}
-        <HelpTip type={helptip} />
-      </p>
+        <span
+          css={`
+            padding-left: 0.2rem;
+            display: inline-block;
+          `}
+        >
+          <HelpTip type={helptip} />
+        </span>
+      </div>
       <span
         css={`
           ${textStyle('title2')};
--- a/src/components/Garden/Metrics.js
+++ b/src/components/Garden/Metrics.js
@@ -14,7 +14,6 @@ const Metrics = React.memo(function Metrics({
   totalSupply,
 }) {
   const { layoutName } = useLayout()
-  const compactMode = layoutName === 'small'
   const currency = {
     name: 'USD',
     symbol: '$',
--- a/src/components/Garden/Metrics.js
+++ b/src/components/Garden/Metrics.js
@@ -14,7 +14,6 @@ const Metrics = React.memo(function Metrics({
   totalSupply,
 }) {
   const { layoutName } = useLayout()
-  // const theme = useTheme()
   const compactMode = layoutName === 'small'
   const currency = {
     name: 'USD',
--- a/src/components/Garden/Metrics.js
+++ b/src/components/Garden/Metrics.js
@@ -29,26 +29,28 @@ const Metrics = React.memo(function Metrics({
           justify-content: space-around;
         `}
       >
-        <div
-          css={`
-            display: flex;
-            align-items: center;
-            margin-bottom: ${(compactMode ? 2 : 0) * GU}px;
-          `}
-        >
-          <img
-            src={token.logo || defaultTokenLogo}
-            height="60"
-            width="60"
-            alt=""
-            onClick={onExecuteIssuance}
+        {!(layoutName === 'medium') && (
+          <div
             css={`
-              margin-right: ${4 * GU}px;
-              cursor: pointer;
+              display: flex;
+              align-items: center;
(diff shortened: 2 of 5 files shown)


src/components/HelpTip.js
```diff
@@ -14,46 +14,65 @@ const KNOWN_HELP_DESCRIPTIONS = {
...
     <>
-      The <strong>Settlement Offer</strong> is the amount of tokens that you
-      would accept from the proposal submitter in order to cancel the action
-      without raising it to Celeste. This amount, if settled, will be slashed
-      from the proposal submitter's collateral and sent to you so it cannot be
-      greater than the collateral locked for the action but it can be zero.
+      The <strong>Settlement Offer</strong> (denominated in your Garden's token)
+      is the amount of tokens you stand to receive from the proposal author, if
+      he or she chooses to withdraw their proposal before it is raised to
+      Celeste.
```
fabriziovigevani, line 54: Do you think this bit of removed info is confusing?

`This amount, if settled, will be slashed from the proposal submitter's collateral and sent to you so it cannot be greater than the collateral locked for the action but it can be zero.`

I think we should make it clear that the settlement offer amount is taken from the author´s collateral so that users understand why the amount cannot be greater than this collateral amount. Maybe we can rephrase it ?

src/components/HelpTip.js
```diff
@@ -14,46 +14,65 @@ const KNOWN_HELP_DESCRIPTIONS = {
...
+    </>,
+  ],
+  'active-supply': [
+    'Active Supply',
+    <>
+      The <strong>Active Supply</strong> is the amount of tokens currently being
+      used to support proposals.
     </>,
   ],
 }
```
fabriziovigevani, line 78: We can update in the below function: 

```
function HelpTip({ type }) {
  const [name, description] = KNOWN_HELP_DESCRIPTIONS[type]
  return <Help hint={`What is ${name}?`}>{description}</Help>
}
```

Add the `the` to the hint text

src/components/Onboarding/Screens/Apps/AgreementSettings/index.js
```diff
@@ -181,10 +181,10 @@ function AgreementSettings() {
...
             <Fragment>
               Settlement Period
-              <Help hint="What is Settlement Period?">
-                <strong>Settlement Period</strong> is the amount of time the
-                proposal's creator has to either accept a settlement or raise
-                the dispute to{' '}
+              <Help hint="What is the Settlement Period?">
+                The <strong>Settlement Period</strong> is the amount of time the
+                proposal author has to either accept a settlement or raise the
+                dispute to{' '}
```
fabriziovigevani, line 187: Realized that we could make use of the `HelpTip` for all the garden settings as well. 

Will create an issue for it

fabriziovigevani (commented):
🙏 🙏 

Some minor suggestions and comments

src/components/Garden/Home.js
```diff
@@ -121,7 +121,7 @@ const Home = React.memo(function Home() {
                   margin: ${(below('medium') ? 0 : 3) * GU}px;
                 `}
               >
-                {!compactMode && (
+                {layoutName !== 'small' && (
```
0xGabi, line 124: I wonder how we should display this information on mobile. Maybe we can use the [accordion component ](https://ui.aragon.org/accordion/) 🤔

0xGabi: https://user-images.githubusercontent.com/9082013/131398930-2711389a-f7c1-4498-88a2-221ac51b87a6.mov

src/components/Garden/Metrics.js
```diff
@@ -29,68 +29,80 @@ const Metrics = React.memo(function Metrics({
...
             value={commonPool}
             token={token}
             currency={currency}
+            helptip="common-pool"
           />
         </div>
         <div>
           <TokenBalance
-            label="Token Supply"
+            label="Total Supply"
```
0xGabi, line 64: What do you think of check the Garden type, and display the `Wrapped Supply` for BYOT Gardens instead? Or maybe both? Given that CV calculations will be made considering the Wrapped Supply.

src/components/HelpTip.js
0xmiel: I don't think the *why* matters too much?

0xmiel: on this topic, i'm wondering whether we should rethink the default settlement offer...

src/components/Garden/Metrics.js
0xmiel: I don't think wrapped is especially clear tbh

0xmiel: but you raise a good point

src/components/Garden/Home.js
0xmiel: I like this. Let's address in another PR?

src/components/Garden/Metrics.js
0xmiel: Let's think about how to best make this distinction

src/components/HelpTip.js
fabriziovigevani: After reviewing your last PR i changed my mind about this. My main concern was that users know that the amount has a maximum and we clearly state that in the form. I agree that the why is not really important in this context.

src/components/Garden/Metrics.js
fabriziovigevani: Yes, good point. Maybe we could add the token symbol in parenthesis for BYOT gardens?

src/components/HelpTip.js
0xmiel: Will have a think about how we can make this clear

src/components/Garden/Metrics.js
0xmiel: Opened an issue for this https://github.com/1Hive/gardens/issues/323

Going ahead and merging

PR #569, opened 2021-08-28
Feat: add side navigation bar

0xGabi:
https://user-images.githubusercontent.com/9082013/131212868-b7f4a6e0-924d-44a1-8351-973aca9aa962.mov

### DIFF
--- a/src/components/MainView.js
+++ b/src/components/MainView.js
@@ -40,7 +40,7 @@ function MainView({ children }) {
 
   return (
     <div css="display: flex">
-      {connectedGarden && (
+      {connectedGarden && !below('medium') && (
         <Sidebar
           gardens={sidebarGardens}
           activeGarden={connectedGarden.address}
--- a/src/components/MainView.js
+++ b/src/components/MainView.js
@@ -1,56 +1,95 @@
-import React from 'react'
+import React, { useCallback, useEffect, useState } from 'react'
+import { useHistory } from 'react-router-dom'
 import { GU, Root, ScrollView, useViewport } from '@1hive/1hive-ui'
-
+import { useGardens } from '@/providers/Gardens'
+import defaultGardenLogo from '@assets/defaultGardenLogo.png'
 import Footer from './Garden/Footer'
 import Header from './Header/Header'
 import Layout from './Layout'
+import Sidebar from './Sidebar/Sidebar'
 
 function MainView({ children }) {
   const { below } = useViewport()
+  const history = useHistory()
+  const { connectedGarden, gardens } = useGardens()
+  const [sidebarGardens, setSidebarGardens] = useState([])
+
   const compactMode = below('large')
 
+  const handleSelectGarden = useCallback(
+    path => {
+      history.push(path)
+    },
+    [history]
+  )
+
+  useEffect(() => {
+    setSidebarGardens(
+      gardens
+        .map(garden => {
+          return {
+            name: garden.name,
+            address: garden.address,
+            path: `/garden/${garden.address}`,
+            src: garden.logo || defaultGardenLogo,
+          }
+        })
+        .slice(0, 10)
+    )
+  }, [gardens])
+
   return (
-    <div
-      css={`
-        display: flex;
-        flex-direction: column;
-        height: 100vh;
-      `}
-    >
-      <Root.Provider
+    <div css="display: flex">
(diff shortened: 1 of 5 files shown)


0xGabi:
I think the current changes add a great UX improvement. We can add the mobile look and feel as in the designs on a subsequent PR.

About the pop-over names. I was not sure how would we should implement it. Appreciate feedback on that regard!

src/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,52 @@
...
+        <Link href="#/home" external={false}>
+          <div
+            css={`
+              border-bottom: 1px solid ${theme.border};
+            `}
+          >
+            <img src={gardensLogo} height={40} alt="" />
+          </div>
+        </Link>
+      </div>
```
fabriziovigevani, line 33: The spacing between the logo and the bottom line here looks a bit off:
Screen Shot 2021-08-30 at 17 58 32

Also i think the bottom line shouldn´t be part of the link to the gardens main page.

We can put the div outside and adjusting the spaces would loook something like

```suggestion
     <div
        css={`
          padding: ${1.5 * GU}px ${2 * GU}px;
        `}
      >
        <div
          css={`
            display: flex;
            padding-bottom: ${1.5 * GU}px;
            border-bottom: 1px solid ${theme.border};
          `}
        >
          <Link
            href="#/home"
            external={false}
            css={`
              display: block;
            `}
          >
            <img
              src={gardensLogo}
              height={40}
              alt=""
              css={`
                display: block;
              `}
            />
          </Link>
        </div>
      </div>
```

src/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,52 @@
...
+      <nav>
+        <ul>
+          {gardens.map((garden, i) => (
+            <MenuItem
+              key={garden.path}
```
fabriziovigevani, line 38: What do you think of using the `garden.address` instead?

```suggestion
              key={garden.address}
```

src/components/MainView.js
```diff
@@ -1,56 +1,95 @@
...
+  )
+
+  useEffect(() => {
+    setSidebarGardens(
+      gardens
+        .map(garden => {
+          return {
+            name: garden.name,
+            address: garden.address,
+            path: `/garden/${garden.address}`,
```
fabriziovigevani, line 33: We missed the `#`
```suggestion
            path: `#/garden/${garden.address}`,
```

src/components/Sidebar/MenuItem.js
```diff
@@ -0,0 +1,38 @@
...
+import { Link, GU } from '@1hive/1hive-ui'
+
+function MenuItem({ active, path, name, src, onOpen }) {
+  const handleClick = useCallback(
+    event => {
+      event.preventDefault()
+      onOpen(path)
+    },
+    [onOpen, path]
+  )
```
fabriziovigevani, line 11: With the added `#` to the `garden.path` i think we can remove this
```suggestion
```

fabriziovigevani (commented):
Awesome job!! 🚀 

I general comments, i think we can encapsulate the logic  of the sidebar and use a different approach.
Also left some spacing fixes and general suggestions.

Regarding the garden name we can use the :after pseudo-element with the :hover pseudo-class in the `li` node with the following:

```
${name &&
    `
     &:hover:after {
        content: "${name}";
        background: #2c3437; // This should have the 0.8 opacity
        border-radius: 12px;
        color: white;
        padding: ${0.5 * GU}px ${2 * GU}px;
        position: absolute;
        top: calc(50% - 16px);
        left: 80px;
        z-index: 2;
     }
`}
```

src/components/Sidebar/Sidebar.js
```diff
@@ -1,46 +1,87 @@
...
   const theme = useTheme()
+  const history = useHistory()
+  const { connectedGarden, gardens } = useGardens()
+
+  const handleSelectGarden = useCallback(
+    path => {
+      history.push(path)
+    },
+    [history]
+  )
```
fabriziovigevani, line 19: This can be removed right? Since we are using the `href` with the constructed path in the MenuItem component. Also we can remove the `onOpen` prop

fabriziovigevani (commented):
🚀 
One last comment and LGTM!

PR #578, opened 2021-09-02
Decisions: Update executeVote function params

fabriziovigevani:
Uses new DisputableVoting interface where the executable script is required as param


0xGabi (approved):
LGTM!

PR #572, opened 2021-09-01
Review: Gardens metadata

rperez89:
PR to match the look and feel of the gardens metadata: <https://www.figma.com/file/ySHlXMwz5elSXKI3feisJU/Gardens-wizard-creation-UI?node-id=495%3A3797>

PR #582, opened 2021-09-02
Onboarding: review aggrement screen

0xGabi:
Builds on #572 
Screen Shot 2021-09-02 at 8 12 20 PM


rperez89 (changes requested):
awesome gabi, we have missing one small thing from the desing is Content/Convenant not just covenant <https://www.figma.com/file/ySHlXMwz5elSXKI3feisJU/Gardens-wizard-creation-UI?node-id=495%3A6461> 

also i am a bit worried about the differences that we are having with the Field labels between pages, take a look to the Metadata ones and this ones.

Screen Shot 2021-09-02 at 9 11 26 PM

0xGabi:
About the `Content/Convenant` wording, I know the designs were like that. But I think it is better to just leave `Covenant`.

I review the `Field` component and we should use this style to have the same style:
```
${textStyle('label2')};

PR #585, opened 2021-09-03
Sort proposals by newest after proposal created

fabriziovigevani:
@0xmiel proposed to sort proposals by newest after a proposal is created so authors of proposals can immediately go and check the proposal without having to do any extra filtering by themselves.

### DIFF
--- a/src/components/Garden/Home.js
+++ b/src/components/Garden/Home.js
@@ -74,6 +74,10 @@ const Home = React.memo(function Home() {
     handleShowModal('unwrap')
   }, [handleShowModal])
 
+  const handleProposalCreated = useCallback(() => {
+    filters.ranking.onChange(1)
+  }, [filters])
+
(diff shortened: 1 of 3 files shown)


src/components/Garden/Home.js
```diff
@@ -74,6 +74,10 @@ const Home = React.memo(function Home() {
     handleShowModal('unwrap')
   }, [handleShowModal])
 
+  const handleProposalCreated = useCallback(() => {
+    filters.ranking.onChange(1)
+  }, [filters])
```
fabriziovigevani, line 79: I´m not totally convinced of this approach, plus we are kind or harcoding the "new" ranking option, but couldn´t come up with a better solution. We can consider a refactor of the filters api so that we can achieve this in a more abstracted way.

0xmiel:
Going ahead and merging

PR #587, opened 2021-09-04
Update transaction messages

0xmiel:
- change `mined` to `processed by network`: clearer and more accurate
- on success, make it clear that one may need to wait a few seconds for changes to be reflected in the UI

### DIFF
--- a/src/components/Stepper/stepper-descriptions.js
+++ b/src/components/Stepper/stepper-descriptions.js
@@ -9,8 +9,10 @@ import {
 export const TRANSACTION_SIGNING_DESC = {
   [STEP_WAITING]: 'Waiting for signature',
   [STEP_PROMPTING]: 'Waiting for signature',
-  [STEP_WORKING]: 'Hang tight. Your transaction is being processed by the network...',
-  [STEP_SUCCESS]: 'Your transaction has successfully been processed! You might need to wait a few seconds for the UI to update',
+  [STEP_WORKING]:
+    'Hang tight. Your transaction is being processed by the network...',
+  [STEP_SUCCESS]:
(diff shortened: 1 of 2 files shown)


src/components/Stepper/stepper-descriptions.js
```diff
@@ -9,8 +9,10 @@ import {
 export const TRANSACTION_SIGNING_DESC = {
   [STEP_WAITING]: 'Waiting for signature',
   [STEP_PROMPTING]: 'Waiting for signature',
-  [STEP_WORKING]: 'Please be patient while your transaction is being mined',
-  [STEP_SUCCESS]: 'Thank you! Your transaction has been successfully mined',
+  [STEP_WORKING]:
+    'Hang tight. Your transaction is being processed by the network...',
+  [STEP_SUCCESS]:
+    'Your transaction has successfully been processed! You might need to wait a few seconds for the UI to update',
```
0xGabi, line 15: I am worried the text might be too long for a 2 seconds period. And the user may have the feeling that they loose something important.

0xGabi: Other than this looks good!

0xmiel: Hmmm, maybe we need to rethink the flow here... Perhaps these modals shouldn'tclose on success.

When one executes a trade on Uniswap, the modal stays open, but it displays a `close` button. When you get used to it, you don't even notice

0xGabi: Yes, we can have a close button for now until we add notifications: https://github.com/1Hive/gardens/issues/330

PR #588, opened 2021-09-04
Sidebar and Header changes

0xmiel:
- Only show connected garden in sidebar (ideally we should show all signed covenant gardens)
- Make sure Covenant is clearly visible in the header on small screens (too easy to miss the footer bar)

### DIFF
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -11,7 +11,7 @@ function Sidebar() {
 
   const gardenData = useMemo(
     () =>
-      gardens.slice(0, 10).map(garden => {
+      gardens.slice(0, 100).map(garden => {
         return {
           name: garden.name,
           address: garden.address,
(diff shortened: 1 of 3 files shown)


src/components/Sidebar/Sidebar.js
```diff
@@ -11,7 +11,7 @@ function Sidebar() {
 
   const gardenData = useMemo(
     () =>
-      gardens.slice(0, 10).map(garden => {
+      gardens.slice(0, 100).map(garden => {
```
fabriziovigevani, line 14: Following previous comment (we´ll need to add the `connectedGarden` as dependency of `useMemo`): 

Also what´s the rationale for increasing the sliced size?

```suggestion
      gardens
        .filter(garden => garden.address === connectedGarden.address)
        .slice(0, 100)
        .map(garden => {
```

fabriziovigevani (commented):
Nice! One small correction 🙏

PR #593, opened 2021-09-05
Feat: use modal flow when create a garden with account disconnected

0xGabi:
Reuse logic from #522 

https://user-images.githubusercontent.com/9082013/132120148-c59c8373-53fb-4c5e-9b11-01727b99db19.mp4


0xGabi:
Include with #522

PR #594, opened 2021-09-05
Fix: handle challenge view on cv visuals

0xGabi:


src/components/Garden/ConvictionVisuals.js
```diff
@@ -14,6 +14,7 @@ import styled from 'styled-components'
 import LineChart from './ModifiedLineChart'
 import SummaryBar from './SummaryBar'
 
+import challengeIconSvg from '@assets/challenge-icon.svg'
```
fabriziovigevani, line 17: Could we move it below after lib and utils imports?

fabriziovigevani (commented):
Since we now introduced a 3rd possible state would like to reestructure a bit the `Outcome` component.

We could make the `Outcome` component accept the `icon` and `color` props and make the outer component responsible for passing the right props.

This would mean that we create a `NegativeOutcome` and `ChallengedOutcome` component.

Also i think the right color for the challenge outcome should be `#F5A623`

WDYT?

PR #597, opened 2021-09-07
Fix: bug with challenge propsal screen

0xGabi:
Fix https://github.com/1Hive/gardens/issues/259

As I understand the `challengerArbitratorFee` could be null and that's what cause the bug: https://github.com/1Hive/honey-pot/blob/gardens/src/utils/data-utils.js#L75

### DIFF
--- a/src/components/Garden/DisputeFees.js
+++ b/src/components/Garden/DisputeFees.js
@@ -50,9 +50,9 @@ function DisputeFees({ proposal }) {
             <div>
               {formatTokenAmount(
                 fees.amount,
-                proposal.challengerArbitratorFee.tokenDecimals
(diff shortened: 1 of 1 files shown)


src/components/Garden/DisputeFees.js
```diff
@@ -50,9 +50,9 @@ function DisputeFees({ proposal }) {
             <div>
```
fabriziovigevani, line 50: Let´s maybe conditionally render this div if `proposal.challengerArbitratorFee` exists?

fabriziovigevani:
Good catch, yeah i think we are in a race condition situation. The proposal status was picked up, but the challengeArbitratorFee  wasn´t yet.

PR #603, opened 2021-09-07
MultiModal: Close button

rperez89:
Since we added more info to the modals we decided to remove the auto close feature (commented the bite of code just in case we want it again when we introduce the activity)


fabriziovigevani (approved):
LGTM!

PR #612, opened 2021-09-12
UX: include a few syle and fixes

0xGabi:


src/components/Garden/Footer.js
```diff
@@ -60,17 +60,18 @@ const defaultFooterData = {
...
 
   useEffect(() => {
     if (connectedGarden) {
-      const { links, logo, token, wrappableToken } = connectedGarden
+      // eslint-disable-next-line camelcase
+      const { links, logo_type, token, wrappableToken } = connectedGarden
       setFooterData({
         links,
-        logo,
+        logo: logo_type,
```
fabriziovigevani, line 74: Where has been decided to use the logo type? Doesn´t it look too big for a footer?

0xGabi: Fiore told me she design using the logo_type in the footer. It will only show on tablet and desktop sizes. So not worried to much about the size. We can consider to reduce the scale.

PR #627, opened 2021-09-17
Proposal footer: Needed tokens fix

fabriziovigevani:
Only display needed tokens when value is `> 0`


rperez89 (approved):
LGTM! good catch

PR #629, opened 2021-09-19
Fixed style for long title proposals

Harthor:


Harthor:
There is so many commits because of fetching 1hive's repo, gonna close this one and open a new one that looks more clear

PR #630, opened 2021-09-19
fixed style for long title proposals

Harthor:


src/components/Garden/Feed/ProposalDescription.js
```diff
@@ -14,6 +14,7 @@ function ProposalDescription({ proposal, onSelectProposal }) {
         margin-bottom: ${3 * GU}px;
         ${textStyle('body1')};
         text-decoration: underline;
+        overflow-wrap: anywhere;
```
fabriziovigevani, line 17: What do you think of also giving the styles to limit the description lines to 3?
```suggestion
        overflow-wrap: anywhere;
        display: -webkit-box;
        -webkit-line-clamp: 3;
        -webkit-box-orient: vertical;
        overflow: hidden;
```

fabriziovigevani (commented):
🙌

src/components/Garden/Feed/ProposalDescription.js
Harthor: > What do you think of also giving the styles to limit the description lines to 3?

the code works correctly, (obviously), i don't have a strong opinion about it, could some proposals need a longer title?

fabriziovigevani: Yeah i mean, if proposals do have a very long title, it would be nice to limit it to just 3 lines, so it doesn{t take a lot of space in the feed. Does that make sense?

Harthor: Yeah, looks great!

fabriziovigevani (approved):
Awesome! Thanks @Harthor

PR #631, opened 2021-09-19
Unipool rewards

fabriziovigevani:
UI for displaying connected account´s accrued rewards and button to claim the
Screen Shot 2021-09-19 at 17 06 52
m.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "3box": "^1.19.1",
     "@1hive/1hive-ui": "1.0.11",
-    "@1hive/connect-gardens": "^0.1.2",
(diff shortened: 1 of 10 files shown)


0xGabi (approved):
Well done! Looking great 🙌

rperez89:
@fabriziovigevani just one cosmetic comment, what do you think if we center the earn rewards section and remove the vertical line from it so the rewards takes the whole down section of the component instead of being on the right and leaving big blank space on the left?

0xmiel:
not sure. I think it's nice this way because it makes it clear that the earned rewards are directly related to the gBright - i'm also not convinced it would look good centered.

0xmiel:
i don't find it awkward tbh

PR #632, opened 2021-09-21
Activity: transactions activities

rperez89:
Screen Shot 2021-09-20 at 10 08 19 PM

Couple of comments, we have a couple of trxs that we are not handling with the modal flow and i was thinking to create a separate pr to change that in order to have all working in the same way.
And the other one is that probably we are going to need the icons at some point or at least if we can have a simple transaction icon one might be enough


src/components/Header/Header.js
```diff
@@ -125,6 +126,17 @@ function Header() {
                 <BalanceModule />
               </>
             )}
+            {account && (
+              <div
```
fabriziovigevani, line 130: Let´s give the parent div a `height: 100%`

So that the :focus background takes the whole height.
Also the popover content should display at the same line where the header ends, not in the middle.

https://user-images.githubusercontent.com/22663232/134428890-3149a51a-9249-487a-9594-e1c8b746d295.mov

src/components/Activity/ActivityButton.js
```diff
@@ -0,0 +1,142 @@
...
+        `}
+      >
+        <ButtonIcon
+          element="div"
+          onClick={handleToggle}
+          css={`
+            height: 100%;
+            width: 100%;
+            border-radius: 0;
+            padding-left: ${0.5 * GU}px;
```
fabriziovigevani, line 63: When clicking the button, seems it´s not fully centered

```suggestion
            padding-left: ${0.75 * GU}px;
```

src/radspec.js
```diff
@@ -0,0 +1,61 @@
+import actions from './actions/garden-action-types'
+
+export default {
+  [actions.ADD_FUNDS]: () => {
+    return `Add funds to deposit manager`
+  },
+  [actions.APPROVE_TOKEN]: ({ symbol }) => {
+    return `Approve ${symbol}`
```
fabriziovigevani, line 8:
```suggestion
  [actions.APPROVE_TOKEN]: ({ tokenSymbol }) => {
    return `Approve ${tokenSymbol}`
```

fabriziovigevani (commented):
Awesome work Ro! 

Left minor suggestions on imports and general corrections.

One thing that was wondering is if we should put more data on the radspec descriptions for certain actions..

For example for token approval, or deposited funds on deposit manager, also state the token amount.

Otherwise the descriptions on the activity don´t really tell much and we could only have the activity title.
Screen Shot 2021-09-22 at 17 36 24

fabriziovigevani:
On another note.

What do you think if we set the tx description to take 2 lines on the Stepper so that it aligns no matter what the txs descriptions length is?

Screen Shot 2021-09-22 at 17 22 46
.

rperez89:
@fabriziovigevani about adding more data to the description i wouldn't recommend to do that unless we place the radspec into another property different than description if not we are also going to see that big description on the modal

like approve 44 HNYT

Screen Shot 2021-09-23 at 8 39 34 PM

fabriziovigevani:
Yeah that´s true, let´s keep as is and think of a way to add more detailed descriptions on the activity item later on.

src/components/Activity/ActivityList.js
```diff
@@ -0,0 +1,145 @@
...
+  textStyle,
+  springs,
+  useTheme,
+  useViewport,
+} from '@1hive/1hive-ui'
+import ActivityItem from './ActivityItem'
+import { useActivity } from '@providers/ActivityProvider'
+
+import { ACTIVITY_STATUS_PENDING } from './activity-statuses'
+import noDataSvg from '@assets/noData.svg'
```
fabriziovigevani, line 15: Could we add a TODO to update this image? Since it´s the one from celeste.

src/radspec.js
```diff
@@ -0,0 +1,61 @@
...
+    return `Create new signalling proposal`
+  },
+  [actions.RESOLVE_ACTION]: ({ disputeId }) => {
+    return `Resolve dispute: ${disputeId}`
+  },
+  [actions.SETTLE_ACTION]: ({ actionId }) => {
+    return `Accept settlement for action: ${actionId}`
+  },
+  [actions.SIGN_AGREEMENT]: () => {
+    return `Sign Agreement`
```
fabriziovigevani, line 41:
```suggestion
    return `Sign Covenant`
```

src/radspec.js
```diff
@@ -0,0 +1,61 @@
...
+    return `Execute decision: ${voteId}`
+  },
+  [actions.EXECUTE_PROPOSAL]: ({ proposalId }) => {
+    return `Execute proposal: ${proposalId}`
+  },
+  [actions.NEW_PROPOSAL]: () => {
+    return `Create new funding proposal`
+  },
+  [actions.NEW_SIGNALING_PROPOSAL]: () => {
+    return `Create new signalling proposal`
```
fabriziovigevani, line 32:
```suggestion
    return `Create new signaling proposal`
```

src/radspec.js
```diff
@@ -0,0 +1,61 @@
...
+  },
+  [actions.RESOLVE_ACTION]: ({ disputeId }) => {
+    return `Resolve dispute: ${disputeId}`
+  },
+  [actions.SETTLE_ACTION]: ({ actionId }) => {
+    return `Accept settlement for action: ${actionId}`
```
fabriziovigevani, line 38:
```suggestion
    return `Settle action: ${actionId}`
```

fabriziovigevani (approved):
Let´s fix this last radspec corrections and LGTM!

PR #642, opened 2021-10-14
Onboarding: Review information screen

fabriziovigevani:
closes https://github.com/1Hive/gardens/issues/151

### DIFF
--- a/src/components/Onboarding/Screens/ReviewInformation.js
+++ b/src/components/Onboarding/Screens/ReviewInformation.js
@@ -198,6 +198,7 @@ function ReviewGardenProfile() {
 }
 /// /////// TOKENOMICS //////////
 function ReviewGardenTokenomics() {
+  const { config } = useOnboardingState()
   return (
     <div
       css={`
@@ -208,8 +209,12 @@ function ReviewGardenTokenomics() {
       <ReviewTokens />
       <LineBreak />
       <ReviewHoneyswapLiquidity />
-      <LineBreak />
-      <ReviewIssuance />
+      {config.garden.type === NATIVE_TYPE && (
+        <>
+          <LineBreak />
+          <ReviewIssuance />
+        </>
+      )}
     </div>
   )
 }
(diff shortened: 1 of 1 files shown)


src/components/Onboarding/Screens/ReviewInformation.js
```diff
@@ -19,4 +63,544 @@ function ReviewInformation({ title }) {
...
+      css={`
+        padding: ${5 * GU}px ${7 * GU}px;
+        width: 100%;
+      `}
+    >
+      <ReviewTokens />
+      <LineBreak />
+      <ReviewHoneyswapLiquidity />
+      <LineBreak />
+      <ReviewIssuance />
```
0xGabi, line 212: We need to check if we are on a Pre-existing token Garden then do not show this.

src/components/Onboarding/Screens/ReviewInformation.js
```diff
@@ -19,4 +63,544 @@ function ReviewInformation({ title }) {
...
+            label="Token address"
+            value={
+              <div>
+                {config.tokens.address}{' '}
+                <span>({config.tokens.existingTokenSymbol})</span>
+              </div>
+            }
+          />
+        )}
+        <Field label="Token name" value={config.tokens.name} />
```
0xGabi, line 253: I notice that the Token address and name are super close without padding. What do you think of moving the Token address on the first row? And then the Token name and symbol on a second row.

fabriziovigevani: Do you think it could help to have the token address in an `IdentityBadge` ?

fabriziovigevani:
Pushed the review changes

There a couple of things yet to do regarding styling the Accordion component but will need to be updated at UI kit level: 

- Modify Accordion item content background
- Make the left border indicator when an item is opened, to follow the container radius for first and last items

src/components/Onboarding/Screens/ReviewInformation.js
0xGabi: I don't think is super relevant. Only to make it easy for the user to click the block explorer and check is the right one.

0xGabi (approved):

PR #645, opened 2021-10-19
Onboarding: Add option to use Gnosis Safe.

rperez89:
Setting for now the second array item on the new trxone signature to zero_addr until we change the ui to reflect this change


rperez89 (commented):
Fabri just an small comment, wondering if token settings  is the right place to do it since is nothing related to a token maybe inside a vault settings?

fabriziovigevani:
@rperez89 Added to the Onboarding the option to set the Gnosis Safe address

Do you think this is the right place to add it?

**Preview**
Screen Shot 2021-10-21 at 21 24 40

Screen Shot 2021-10-21 at 21 24 50

0xGabi (approved):
Well done!

PR #650, opened 2021-10-20
Evm executor

rperez89:


src/components/Garden/Preferences/GlobalPreferences.js
```diff
@@ -43,15 +49,18 @@ function GlobalPreferences({ compact, onClose, onNavigation, sectionIndex }) {
...
-        <React.Fragment>
-          <Tabs
-            items={VALUES}
-            onChange={onNavigation}
-            selected={sectionIndex}
-          />
-
-          {sectionIndex === NETWORK_INDEX && <AppsAddresses />}
-        </React.Fragment>
+        <Root.Provider>
```
fabriziovigevani, line 52: Is this Root provider for the multi modal?

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+import CreateDecisionScreens from '../ModalFlows/CreateDecisionScreens/CreateDecisionScreens'
+
+import { useGardens } from '@/providers/Gardens'
+import { useWallet } from '@/providers/Wallet'
+
+import { SHORTENED_APPS_NAMES } from '@utils/app-utils'
+// import { asciiToHex } from '@utils/web3-utils'
+
+import actions from '../../../actions/garden-action-types'
+import radspec from '../../../radspec'
```
fabriziovigevani, line 23:
```suggestion
import actions from '@/actions/garden-action-types'
import radspec from '@/radspec'
```

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+  )
+  const [evmcrispr, setEvmcrispr] = useState(null)
+  const [installedApps, setInstalledApps] = useState([])
+  const [selectedApp, setSelectedApp] = useState(null)
+  const [selectedFunction, setSelectedFunction] = useState(null)
+  const [parameters, setParameters] = useState([])
+
+  useEffect(() => {
+    async function getEvmCrispr() {
+      if (!connectedGarden) {
```
fabriziovigevani, line 39: Let´s also add the `!account` condition because it errors on ethers.getSigner otherwise.

Also wondering if the `!connectedGarden` condition is necessary in this case? When getting to this page doesn´t it mean we are indeed in a garden?
```suggestion
      if (!connectedGarden || !account) {
```

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+    setInstalledApps(apps)
+  }, [evmcrispr])
+
+  const shortenedAppsNames = installedApps.map(appName => {
+    const dotIndex = appName.indexOf('.')
+    return (
+      SHORTENED_APPS_NAMES.get(
+        dotIndex > 0
+          ? appName.substr(0, appName.indexOf('.'))
+          : appName.slice(0, -2)
```
fabriziovigevani, line 66: Not sure i understand this. In what case we need to remove the last 2 chars of the app name?

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+      setEvmcrispr(crispr)
+    }
+    getEvmCrispr()
+  }, [connectedGarden, ethers])
+
+  useEffect(() => {
+    if (!evmcrispr) {
+      return
+    }
+    const apps = evmcrispr.apps()
```
fabriziovigevani, line 56: Since this is not something returned async, couldn´t we just use a `useMemo` here? Like we do for the functions for example.

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+        dotIndex > 0
+          ? appName.substr(0, appName.indexOf('.'))
+          : appName.slice(0, -2)
+      ) || appName.slice(0, -2)
+    )
+  })
+
+  const functionList = useMemo(() => {
+    if (!evmcrispr || shortenedAppsNames?.length === 0 || !selectedApp) {
+      return
```
fabriziovigevani, line 73: For useMemos let´s always return a value  (`null`or `[]`)

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+
+    const appName = installedApps[selectedApp]
+
+    const appFunctions = Object.getOwnPropertyNames(evmcrispr.call(appName))
+    return appFunctions
+  }, [evmcrispr, selectedApp, installedApps, shortenedAppsNames])
+
+  const requiredParameters = useMemo(() => {
+    if (!evmcrispr || !selectedApp || !selectedFunction) {
+      return
```
fabriziovigevani, line 84: Same here

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+        >
+          You must connect your account in order to create a decision.
+        </Info>
+      )}
+      {selectedFunction && (
+        <Button
+          disabled={!account}
+          mode="strong"
+          wide
+          onClick={() => setCreateDecisionModalVisible(true)}
```
fabriziovigevani, line 188: Could we create callbacks both for here and the multimodal `onClose`?

src/utils/app-utils.js
```diff
@@ -16,6 +16,19 @@ export const SHORTENED_APPS_NAMES = new Map([
   ['wrappable-hooked-token-manager', 'token-manager'],
 ])
 
+export const FULL_NAME_FROM_SHORTENED = new Map([
```
fabriziovigevani, line 19: I don´t see this being used, should we remove it?

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+    })
+  }, [])
+
+  const handleOnCreateIntent = useCallback(async () => {
+    const description = radspec[actions.NEW_DECISION]()
+    const type = actions.NEW_DECISION
+    console.log(
+      'functionList[selectedFunction] ',
+      functionList[selectedFunction]
+    )
```
fabriziovigevani:
```suggestion
```

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+    const intent = await evmcrispr.encode(
+      [
+        evmcrispr
+          .call(installedApps[selectedApp])
+          [functionList[selectedFunction]](...parameters),
+      ],
+      ['disputable-voting'],
+      // TODO: just for now that for some reason the radspec description on the card is not working, after fixed we can ask the user for enter some forum post related to why the decision is being created
+      // { context: asciiToHex(functionList[selectedFunction]) }
+      // having some issue on the lib when passing the function that need to check with david
```
fabriziovigevani, line 117: Let´s discuss offline about this.

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -0,0 +1,207 @@
...
+    setParameters(prevState => {
+      const newArray = [...prevState]
+      newArray[index] = newValue
+      return newArray
+    })
+  }, [])
+
+  const handleOnCreateIntent = useCallback(async () => {
+    const description = radspec[actions.NEW_DECISION]()
+    const type = actions.NEW_DECISION
```
fabriziovigevani, line 107: Let´s perhaps add a TODO: here to move this to the file of `useActions` as i think we should have them all there once we change the rest to use the `evmscripr` (if we decide to do so)

fabriziovigevani (commented):
Awesome job Ro!
Looks really good.

The concern i have is that this won´t work for all actions across all the apps.
For example we expose all the functions from conviction voting app and assuming that all will go through `disputable-voting` Not sure if we should filter them by the only ones that are governed by the voting app but we don´t have a very good way to do that dynamically based on the permissions i think.

But i think we should discuss it.

Moreover we should probably create an issue to make all actions to use `evmscripr`. And i think we could also create a provider which exposes all actions across the garden and move the ones we have in `useActions` to said provider.

What do we think about this? cc. @0xGabi

src/components/Garden/Preferences/EVMExecutor.js
rperez89: is to be sure that the `connectedGarden.address,` is going to have a value

src/components/Garden/Preferences/EVMExecutor.js
rperez89: because the crispr returns the app plus two numbers like 01 that represents the instance in case there are more than one instance

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -47,30 +46,35 @@ function EVMExecutor() {
...
+  }, [account, connectedGarden, ethers])
 
-  useEffect(() => {
+  // useEffect(() => {
+  //   if (!evmcrispr) {
+  //     return
+  //   }
+  //   const apps = evmcrispr.apps()
+  //   setInstalledApps(apps)
+  // }, [evmcrispr])
```
fabriziovigevani, line 57: Should we remove this?

Same for installedApps state

fabriziovigevani (approved):
🚀

src/components/Garden/ModalFlows/CreateDecisionScreens/CreateDecisionScreens.js
```diff
@@ -0,0 +1,76 @@
...
+      const intent = await onCreateTransaction()
+      setTransactions(intent)
+      onComplete()
+    },
+    [onCreateTransaction]
+  )
+
+  const screens = useMemo(
+    () =>
+      stakingLoading
```
0xGabi, line 37: @rperez89 I was doing a review of the whole EVM executor. Shouldn't be the `loading` variable here instead of the `stakingLoading`?

PR #657, opened 2021-10-26
Sidebar changes

PJColombo:
This PR solves both issue 336 and 333.

It displays the sidebar on the profile page, makes it scrollable and changes its content by replacing the connected garden with the user signed gardens.
 

output

### DIFF
--- a/src/components/Sidebar/MenuItem.js
+++ b/src/components/Sidebar/MenuItem.js
@@ -8,7 +8,7 @@ function MenuItem({ active, path, label, src }) {
     <li
       css={`
         position: relative;
-        background: ${active ? 'rgb(227, 240, 213)' : theme.surface};
+        background: ${active ? 'rgb(212 251 216)' : theme.surface};
         height: ${8.5 * GU}px;
         width: 100%;
         ${label &&
@@ -19,7 +19,6 @@ function MenuItem({ active, path, label, src }) {
               border-radius: 12px;
               color: white;
               padding: ${0.5 * GU}px ${2 * GU}px;
-              position: absolute;
               top: calc(50% - 16px);
               left: 80px;
           }
--- a/src/components/Sidebar/MenuItem.js
+++ b/src/components/Sidebar/MenuItem.js
@@ -8,7 +8,7 @@ function MenuItem({ active, path, name, src }) {
     <li
       css={`
         position: relative;
-        background: ${active ? 'rgb(249, 249, 248)' : theme.surface};
+        background: ${active ? 'rgb(227, 240, 213)' : theme.surface};
         height: ${8.5 * GU}px;
         ${name &&
           ` &:hover:after {
@@ -33,7 +33,7 @@ function MenuItem({ active, path, name, src }) {
             bottom: 0;
             left: 0;
             width: 4px;
-            background: rgb(141, 233, 149);
+            background: rgb(62, 207, 75);
           `}
         />
       )}
@@ -41,6 +41,10 @@ function MenuItem({ active, path, name, src }) {
         css={`
           display: flex;
           padding: ${1.5 * GU}px;
+          ${!active &&
+            `&:hover {
+            background: rgb(246, 246, 247);
+          }`}
         `}
       >
         <Link
--- a/src/components/Sidebar/MenuItem.js
+++ b/src/components/Sidebar/MenuItem.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Link, GU, useTheme } from '@1hive/1hive-ui'
 
-function MenuItem({ active, path, name, src }) {
+function MenuItem({ active, path, label, src }) {
   const theme = useTheme()
 
   return (
@@ -10,9 +10,11 @@ function MenuItem({ active, path, name, src }) {
         position: relative;
         background: ${active ? 'rgb(227, 240, 213)' : theme.surface};
         height: ${8.5 * GU}px;
-        ${name &&
+        width: 100%;
+        ${label &&
           ` &:hover:after {
-              content: "${name}";
+              position: fixed;
+              content: "${label}";
               background: rgba(44, 52, 55, 0.8);
               border-radius: 12px;
               color: white;
@@ -20,7 +22,6 @@ function MenuItem({ active, path, name, src }) {
               position: absolute;
               top: calc(50% - 16px);
               left: 80px;
-              z-index: 2;
           }
         `}
       `}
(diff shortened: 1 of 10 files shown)


src/components/Sidebar/Sidebar.js
```diff
@@ -65,18 +85,55 @@ function Sidebar() {
...
+              css={`
+                margin-top: ${6 * GU}px;
+                margin-left: ${2 * GU}px;
+              `}
+            >
+              <LoadingRing mode="half-circle" />
+            </div>
+          ) : (
+            <ul>
+              {trail.map((style, index) => {
```
fabriziovigevani, line 121: Nice :fire:

src/hooks/useUser.js
```diff
@@ -6,29 +6,39 @@ import { transformUserData } from '@utils/data-utils'
...
         const user = await getUser(
           { network: getNetwork().chainId },
           { id: address.toLowerCase() }
         )
+        setLoading(false)
         if (mounted()) {
           setUser(transformUserData(user))
         }
       } catch (err) {
+        setLoading(false)
```
fabriziovigevani, line 35: We could update the loading state after the try catch once instead of in each block.

src/components/Sidebar/Sidebar.js
```diff
@@ -1,33 +1,54 @@
...
-import { GU, useTheme, Link } from '@1hive/1hive-ui'
+import { useTrail, animated } from 'react-spring'
+import { GU, Link, useTheme } from '@1hive/1hive-ui'
 import gardensLogo from '@assets/gardensLogoMark.svg'
 import defaultGardenLogo from '@assets/defaultGardenLogo.png'
-import MenuItem from './MenuItem'
+import useUser from '@/hooks/useUser'
 import { useGardens } from '@/providers/Gardens'
+import { useWallet } from '@/providers/Wallet'
+import { addressesEqual } from '@/utils/web3-utils'
```
fabriziovigevani, line 9: Nit pick:

```suggestion
import useUser from '@hooks/useUser'
import { useGardens } from '@providers/Gardens'
import { useWallet } from '@providers/Wallet'
import { addressesEqual } from '@utils/web3-utils'
```

src/components/Sidebar/Sidebar.js
```diff
@@ -1,33 +1,54 @@
...
+      const { address, name, logo } = gardens.find(g =>
+        addressesEqual(g.address, gardenSignedAddress)
+      )
+      return {
+        address,
+        name,
+        path: `#/garden/${address}`,
+        src: logo || defaultGardenLogo,
+      }
+    })
```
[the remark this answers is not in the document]
fabriziovigevani: I agree, let´s keep it this way for now and optimize if needed.

fabriziovigevani (approved):
Awesome Paulo 🚀 🚀 

Regarding the reloading of the sidebar it´s caused by the GardensProvider, where if there´s a garden connected then it will return the `cihldren` wrapped in the necessary providers. and if not it will return just the children.  This makes the `Sidebar` be a new render if we move from a page where there´s a connected garden to one where there isn´t.

What i imagine we could do is have a local map cache in the SideBar where we check if the gardens sidebar is present on said cache. If not we construct it with the gardens list and gardensSigned as we do now. 

What do you think?

fabriziovigevani:
> Oh, you're right. We could memoize the sidebar gardens just once by adding a boolean to the dependencies so we don't recalculate them when the gardens dependency changes.

Hmm not sure this would solve the issue as it asumes all gardens will be available when constructing the sidebar gardens for the first time.

If for example we first access the gardens home page, we filter by a garden and then access our profile the issue will happen regardless.

Maybe we would need for the gardens metadata list to be available for the rest of the app to use?

src/components/Sidebar/MenuItem.js
```diff
@@ -1,26 +1,27 @@
...
-function MenuItem({ active, path, name, src }) {
+function MenuItem({ active, path, label, src }) {
   const theme = useTheme()
 
   return (
     <li
       css={`
         position: relative;
-        background: ${active ? 'rgb(249, 249, 248)' : theme.surface};
+        background: ${active ? 'rgb(227, 240, 213)' : theme.surface};
```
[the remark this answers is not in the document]
PJColombo: Yeah, I agree.

PJColombo:
> Hmm not sure this would solve the issue as it asumes all gardens will be available when constructing the sidebar gardens for the first time.
> 
> If for example we first access the gardens home page, we filter by a garden and then access our profile the issue will happen regardless.
> 
> Maybe we would need for the gardens metadata list to be available for the rest of the app to use?

Yeah, you're right. I totally forgot about the fact that we don't have the sidebar on the homepage.

I agree on using the metadata list instead as it contains all the data (name and logo) we need for rendering the sidebar elements.

PJColombo:
> Awesome Paulo rocket rocket
> 
> Regarding the reloading of the sidebar it´s caused by the GardensProvider, where if there´s a garden connected then it will return the `cihldren` wrapped in the necessary providers. If not it will return just the children. This makes the `Sidebar` be a new render if we move from a page where there´s a connected garden to one where there isn´t (e.g. Garden home -> Profile)
> 
> What i imagine we could do is have a local map cache in the Sidebar (key being the account and value the list of gardens to render on the sidebar) where we check if the gardens sidebar is present on said cache. If not we construct it with the gardens list and gardensSigned as we do now.
> 
> What do you think?

I like this solution but what about just caching the current connected account instead?. This way the animation is triggered every time you connect an account.

fabriziovigevani:
> The metadata list could be useful for rendering the sidebar gardens but if I try to go to one that isn't included in the current gardens filtered list it will throw an error anyway.

Correct. What i was suggesting was using the `gardensSigned` form the user entity along with just the gardens metadata list. So we would be exposing `gardens` and `gardensMetadata` lists from the provider. I´m not really a fan of having a list where the content is already present in another list but i think given the case it makes sense as the `gardens` list isn´t actually the whole list of gardens but a filtered one.

This way there´s no need to update the Schema in the subgraph either.

fabriziovigevani:
> I like this solution. What do you think of only caching the current connected account instead ? This way the animation is triggered every time you connect an account.

Are you implying that we remove from the cache the current connected account when it disconnects?

fabriziovigevani:
@PJColombo @0xGabi i did the PR comments that we reviewed. 

Also i created a provider User which holds the user data so that it can be consumed by Agreement when signing new agreements (so we can refetch the user data) and also so that the user data is already loaded when changing between A garden and the Profile screen. This way we don´t need to have the Cache we were discussing. Though the animation will fire when changing between these screens which i think is fine.

LMKWYT

0xGabi (approved):

PR #663, opened 2021-11-01
UX: update the modal sign illustrations

0xGabi:
Fix https://github.com/1Hive/gardens/issues/132

https://user-images.githubusercontent.com/9082013/139708935-7c4da09e-f69e-4653-bf62-7f2676f4a5d1.mov

Should we keep the old HNY logos for failing and confirm actions?


fabriziovigevani:
> Should we keep the old HNY logos for failing and confirm actions?

If we could get the same asset customised for those states i think would be better.

fabriziovigevani:
On a general note and for future PRs. Could we optimize only the necessary assets? Otherwise the files changed get a bit confusing.

0xGabi:
> On a general note and for future PRs. Could we optimize only the necessary assets? Otherwise the files changed get a bit confusing.

Sure, I did it on two separate commits. But I will include the changes in two separate PRs next time.

PR #664, opened 2021-11-01
Chore: add link to home header logo

0xGabi:
Solves https://github.com/1Hive/gardens/issues/364

Instead of including the `WelcomeModal` we add a link to the Gardens landing page.


fabriziovigevani (approved):
LGTM

PR #666, opened 2021-11-01
App(onboarding):remove delegate voting period from advanced settings

0xGabi:
Fix https://github.com/1Hive/gardens/issues/387


fabriziovigevani (approved):
LGTM!

PR #671, opened 2021-11-05
Feat(networks): add support for network switch and allow

0xGabi:
Add support for [EIP-3326](https://eips.ethereum.org/EIPS/eip-3326) and [EIP-3085](https://eips.ethereum.org/EIPS/eip-3085).

### DIFF
--- a/src/networks.js
+++ b/src/networks.js
@@ -4,24 +4,16 @@ import {
   getNetworkName,
 } from '@utils/web3-utils'
 import { getPreferredChain } from './local-settings'
-import { DEFAULT_CHAIN_ID } from './constants'
 import env from './environment'
 
-const ETH_NODE = env('ETH_NODE')
+const XDAI_ETH_NODE = env('XDAI_ETH_NODE')
 
 const networks = {
-  mainnet: {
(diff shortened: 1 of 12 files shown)


src/networks.js
```diff
@@ -65,17 +57,69 @@ const networks = {
...
+      method: 'wallet_switchEthereumChain',
+      params: [{ chainId: chainIdHex }],
+    })
+  } catch (switchError) {
+    // This error code indicates that the chain has not been added to Injected provider.
+    if (switchError.code === 4902) {
+      try {
+        await addEthereumChain()
+      } catch (addError) {
+        // handle "add" error
```
rperez89, line 115: maybe here at least we can throw a console.error for now

src/networks.js
```diff
@@ -65,17 +57,69 @@ const networks = {
...
+      }
+    }
+    // handle other "switch" errors
+    // TODO: handle errors
```
rperez89, line 119: same here

rperez89 (commented):
Awesome work gabi!!

0xGabi:
I address the logs in #672

PR #674, opened 2021-11-09
Support different request token

fabriziovigevani:
closes https://github.com/1Hive/honey-pot/issues/656

- Updated common pool token to be request token
- Rearranges the Metrics component:
Screen Shot 2021-11-09 at 00 21 59
Screen Shot 2021-11-09 at 00 22 25
It now shows wrapped supply for BYOT gardens

- Create proposal: Removed the option to request amount in stable value if request token is already a stable token.

### DIFF
--- a/src/logic/garden-logic.js
+++ b/src/logic/garden-logic.js
@@ -36,7 +36,7 @@ export default function useGardenLogic() {
       value: config?.conviction.totalStaked,
       token: config?.conviction.stakeToken,
     },
-    // mainToken: BYOT ? wrappableToken : token
+    // For BYOT gardens, mainToken will be `wrappableToken`, for native gardens will be `token`
     totalSupply: { value: mainToken.totalSupply, token: mainToken.data },
     // For BYOT we will also display the total suppply of the wrapped token
     totalWrappedSupply: wrappableToken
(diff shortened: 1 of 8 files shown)


src/logic/garden-logic.js
```diff
@@ -16,13 +24,23 @@ export default function useAppLogic() {
...
     loading: loading || !blockHasLoaded,
-    mainToken,
+    priceToken: mainToken.data,
     proposals,
     proposalsFetchedCount,
+    totalActiveTokens: {
+      value: config?.conviction.totalStaked,
+      token: config?.conviction.stakeToken,
+    },
+    // mainToken: BYOT ? wrappableToken : token
```
0xGabi, line 39: Let's remove this line.

0xGabi (approved):
Great work. The metrics and common pool separation make the whole dashboard more approachable for BYOT Gardens.

I will include some small corrections I notice on the helptip wording on a new PR.

src/logic/garden-logic.js
fabriziovigevani: Ah i put this line not as something commented that forgot to remove but for others to consider what the main token is. Perhaps we can change it to plain english ? lol

0xGabi (approved):

PR #677, opened 2021-11-09
Sidebar fixes

PJColombo:
This PR solves the following problems related to the sidebar component:

1. The sidebar garden's hint label positioning.
2. Error handling when one of the signed gardens isn't found in the `gardensMetadata` list.
3. Fetch connected garden from subgraph instead of getting it from the local gardens list.
4. Speed up the  active sidebar garden rendering by getting the address from the URL instead of waiting for the connected garden to be fetched.


src/providers/Gardens.js
```diff
@@ -23,30 +23,26 @@ import { getGardenForumUrl } from '../utils/garden-utils'
 import useGardenFilters from '@/hooks/useGardenFilters'
 import { testNameFilter } from '@/utils/garden-filters-utils'
 import { useDebounce } from '@/hooks/useDebounce'
+import { useMounted } from '@/hooks/useMounted'
 
 import { getVoidedGardensByNetwork } from '../voided-gardens'
+import { getGarden } from '@1hive/connect-gardens/dist/cjs/gardens'
```
0xGabi, line 29: We should update this line to export directly from `@1hive/connect-gardens`.

src/providers/Gardens.js
```diff
@@ -23,30 +23,26 @@ import { getGardenForumUrl } from '../utils/garden-utils'
...
 
 const DAOContext = React.createContext()
 
 export function GardensProvider({ children }) {
   const [queryFilters, filters] = useGardenFilters()
-  const [gardens, gardensMetadata, loading, reload] = useGardensList(
+  const [gardens, gardensMetadata, gardensLoading, reload] = useGardensList(
```
0xGabi, line 35: With the new changes to support multichain. The logic of this file is going to change a bit given that we will use the `preferredNetwork` as chainId instead of `getNetwork().chainId`.

0xGabi (approved):
Well done! 

We have to update the dependency of the connector now that we have published the new version.

src/providers/Gardens.js
PJColombo: Oops, I completely forgot about removing this when testing the connector locally.

0xGabi (approved):

fabriziovigevani:
@PJColombo I added a few changes [here](https://github.com/1Hive/honey-pot/pull/693) where we refactor the logic of the connected garden. We moved down the Conected garden ito its own provider to separate logic sincd it was giving some issues with knowing when we were fetching a garden and it will help in the future with the araognId support.

PR #679, opened 2021-11-10
Chore: fix helptip wordings

0xGabi:


src/components/HelpTip.js
```diff
@@ -64,15 +64,15 @@ const KNOWN_HELP_DESCRIPTIONS = {
...
+      The <strong>Total Supply</strong> is the total amount of tokens in
       circulation, including the common pool.
     </>,
   ],
   'total-wrapped-supply': [
     'Total Wrapped Supply',
     <>
-      The <strong>Total Wrapped Supply</strong> is the the total amount of
-      wrapped tokens in circulation.
+      The <strong>Total Wrapped Supply</strong> is the amount of wrapped tokens
```
fabriziovigevani, line 74: Ah maybe we should update also `Total Wrapped Supply` for just `Wrapped Supply`?

PR #682, opened 2021-11-16
Evm external executor

rperez89:
Couple of considerations:
For now we are just supporting the user to enter the abi in json format.
We need to find a way to show something on the decision vote card maybe add on the context the function and the parameters with is being called?


src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -100,29 +163,79 @@ function EVMExecutor() {
...
+    let intent
+
+    if (interactionType === INTERNAL_INDEX) {
+      intent = await evmcrispr.encode(
+        [
+          evmcrispr
+            .call(installedApps[selectedApp])
+            [functionList[selectedFunction]](...parameters),
+        ],
+        ['disputable-voting'],
```
0xGabi, line 176: Here we should generalize for cases like on `xdai` and `polygon` where the app we install is on the open AragonPM registry. So the `EVMCrispr` expect `disputable-voting.open` instead.

0xGabi (changes requested):
Well done Ro! The whole EVMexecutor setting is an awesome new feature 💪

src/components/Garden/Preferences/EVMExecutor.js
rperez89: nice cathc!

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -57,6 +62,14 @@ function EVMExecutor() {
     getEvmCrispr()
   }, [account, connectedGarden, ethers])
 
+  const forwarderName = useMemo(() => {
+    if (!gardenState || !gardenState.installedApps) {
+      return null
+    }
+    return getAppByName(gardenState.installedApps, env('VOTING_APP_NAME'))
+      .artifact.appName
```
0xGabi, line 70: EVMcrispr do not understand the full `appName`. Let's do `appName.split('.aragonpm.eth')[0]`. To only keep the first part.

0xGabi (changes requested):
I left one request for change. Other than that looks good to me! 💪

0xGabi (approved):

PR #688, opened 2021-11-18
Delegated voting

fabriziovigevani:
closes https://github.com/1Hive/gardens/issues/247

Includes: 

- Delegate voting component to set, change or remove a delegate.
- Profile component that shows a given account´s delegates across all gardens as well as users the account is delegate of across gardens.
- Updated Decision detail UI to vote on behalf of principals which includes:
	- Moved voting actions to use MultiModal
	- Add transaction to vote on behalf of principals if it corresponds.
	-  Update TokenReference to show how much voting power comes from delegated voting
	- Update VoteCasted component to take into account delegated voting power
	- Allow principals to override delegate vote.

**Preview**

1 - Component to show a delegate:

- Empty: 
Untitled

- Delegate set: 
Untitled 2

2 - Component to update a delegate

Screen Shot 2021-11-17 at 22 17 54
 
3 - Profile component that shows a given account´s delegates across all gardens as well as users the account is delegate of across gardens.

Screen Shot 2021-11-17 at 22 16 54

4 - Decision detail updates: 

- Delegate voting (account got chosen by two users to be delegate for)

Screen Shot 2021-11-17 at 22 20 03
 
5 - Adding tx to vote on behalf of principals

Screen Shot 2021-11-17 at 22 20 19

6 - Result after voting: 

Screen Shot 2021-11-17 at 22 38 15

7 - Principals voting information (as you can note, they can override their delegate vote)

First principal
Screen Shot 2021-11-17 at 22 43 58

Second principal
Screen Shot 2021-11-17 at 22 44 14


src/components/Garden/Feed/Delegation.js
```diff
@@ -0,0 +1,160 @@
...
+        onClick={handleViewProfile}
+        css={`
+          margin-bottom: ${1 * GU}px;
+          cursor: pointer;
+        `}
+      >
+        {profile?.image ? (
+          <img
+            src={profile.image}
+            height="72"
```
rperez89, line 111: let's use the GU

src/components/Garden/Feed/RightPanel.js
```diff
@@ -1,91 +1,83 @@
...
+  const styles = useMemo(() => {
+    if (largeMode) {
+      return {
+        display: 'flex',
+        flexDirection: 'column',
+        marginTop: `${3 * GU}px`,
+        position: 'sticky',
+        rowGap: `${3 * GU}px`,
+        top: `${3 * GU}px`,
+        width: '327px',
```
rperez89, line 37: let's use the GU

src/components/Profile/Delegates.js
```diff
@@ -0,0 +1,157 @@
...
+          ${textStyle('label2')};
+          margin-bottom: ${1 * GU}px;
+        `}
+      >
+        {addressesEqual(account, connectedAccount) ? 'My ' : ''}Delegates
+      </h3>
+      {user?.supports && user.supports.length > 0 ? (
+        user.supports.map((support, index) => {
+          if (!support.representative) {
+            return null
```
rperez89, line 64: since we are inside html maybe is  better to return </> instead of  null

rperez89 (commented):
Woooow fabri this is huge!! awesome work man!!!

fioreb:
Here you can find the designs for the Delegate voting section, I did it super simple for the v1 release:
https://www.figma.com/file/OksvE2Qph3TOh8sxBwu7dO/Delegate-voting?node-id=0%3A1

And the new icon for the user menu

src/utils/vote-utils.js
```diff
@@ -22,18 +23,34 @@ export function getAccountCastStake(vote, account) {
     addressesEqual(cast.supporter.user.address, account)
   )
 
-  return userCast?.stake || 0
+  return bigNum(userCast?.stake || 0, 0)
 }
 
-export function getConnectedAccountVote(vote, account) {
+export function getAccountCastDelegatedStake(vote, account) {
+  // Takes into account delegated cast stakes
```
fabriziovigevani, line 30:
```suggestion
  // Takes into account delegated cast stakes (casts done by account where account === caster !== supporter, supporter being caster´s principal)
```

src/components/Garden/ModalFlows/DelegateVotingScreens/DelegateVoting.js
```diff
@@ -0,0 +1,139 @@
...
+            css={`
+              display: flex;
+              align-items: center;
+              column-gap: 8px;
+            `}
+          >
+            <Field
+              label={`Your ${
+                hasRepresentative ? 'new' : ''
+              } delegate´s address`}
```
0xGabi, line 64:
```suggestion
              } delegate's address`}
```

src/components/Garden/DecisionDetail/VoteCasted.js
```diff
@@ -54,15 +61,17 @@ function VoteCasted({ account, accountVote, vote }) {
             margin-bottom: ${0.5 * GU}px;
           `}
           >
-            Your vote was cast successfully
+            {`Your ${
+              addressesEqual(account, caster) ? '' : 'delegate´s'
```
0xGabi, line 65:
```suggestion
              addressesEqual(account, caster) ? '' : 'delegate's'
```

0xGabi (changes requested):
Amazing work Fabri!! This new feature is super valuable and makes the app feature complete. Thanks 🙌

src/components/Garden/DecisionDetail/DecisionDetail.js
```diff
@@ -35,13 +36,13 @@ import { useWallet } from '@providers/Wallet'
...
 import { round, safeDiv } from '@utils/math-utils'
-import { getConnectedAccountVote, getQuorumProgress } from '@utils/vote-utils'
+import { getConnectedAccountCast, getQuorumProgress } from '@utils/vote-utils'
 
 import { PCT_BASE, VOTE_NAY, VOTE_YEA } from '@/constants'
 
 function DecisionDetail({ proposal, actions }) {
   const [modalVisible, setModalVisible] = useState(false)
-  const [modalMode, setModalMode] = useState(null)
+  const [modalData, setModalData] = useState({ mode: '' })
```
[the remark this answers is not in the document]
fabriziovigevani: Ah i did this to store some extra data that will be passed to some of the multimodal screens. For example if you take a look at the `handleVote`You can see that we pass extra data that will be passed to the `VoteOnDecisionScreens`.

src/components/Garden/Feed/Delegation.js
fabriziovigevani: We have agreed in the past that we´ll use GU only for margins, paddings and other spacing styles but not for element dimension.

src/components/Profile/Delegates.js
fabriziovigevani: `null` should be the one to use when we don´t want to return anything https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

src/hooks/useSubscriptions.js
```diff
@@ -104,7 +104,7 @@ export function useProposalSubscription(proposalId, appAddress) {
   const onProposalHandler = useCallback(
     async (err, proposal) => {
       if (err || !proposal) {
-        setLoading(true)
+        setLoading(false)
```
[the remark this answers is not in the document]
fabriziovigevani: Not sure if it was influencing in any way, but going across this file it called my attention that we were doing this. If there´s an error of some kind we should signal that we stop loading.

src/components/Garden/ModalFlows/DelegateVotingScreens/DelegateVoting.js
```diff
@@ -0,0 +1,139 @@
...
+    </div>
+  )
+}
+
+function NewDelegateProfile({ address }) {
+  const profile = useProfile(address)
+
+  if (!profile) {
+    return null
+  }
```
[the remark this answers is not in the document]
fabriziovigevani: In the case of CurrentDelegateProfile there´s no need because we need to show either the address or profile.

In the case of the `NewDelegateProfile`  we already have the new delegate address in the input, so thought it would only make sense to show the IdentityBadge if it would show the profile

src/components/Profile/Delegates.js
```diff
@@ -0,0 +1,157 @@
...
+        border-top: 1px solid ${theme.border};
+      `}
+    >
+      <h3
+        css={`
+          ${textStyle('label2')};
+          margin-bottom: ${1 * GU}px;
+        `}
+      >
+        Delegate for
```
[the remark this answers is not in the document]
fabriziovigevani: Can update but note that in this case since the textStyle is of a label it will all get converted to upper case anyway.

src/radspec.js
```diff
@@ -52,8 +61,13 @@ export default {
...
   },
-  [actions.VOTE_ON_DECISION]: ({ voteId }) => {
-    return `Vote on decision: ${voteId}`
+  [actions.VOTE_ON_DECISION]: ({ voteId, supports }) => {
+    return `Vote ${supports ? 'Yes' : 'No'} on decision: #${voteId}`
+  },
+  [actions.VOTE_ON_BEHALF_OF]: ({ voteId, supports }) => {
+    return `Vote ${
+      supports ? 'Yes' : 'No'
+    } on behalf of principals on decision: #${voteId}`
```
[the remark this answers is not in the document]
fabriziovigevani: Principals refers to the users that are delegating to you,  representatives refer to users you are delegating to (they are your representatives).

0xGabi (approved):

PR #690, opened 2021-11-24
App(network): add support for Polygon

0xGabi:


src/networks.js
```diff
@@ -80,11 +116,12 @@ export function getNetwork(chainId = getPreferredChain()) {
 }
 
 export function getEthersNetwork() {
-  const { type, chainId, ensRegistry } = getNetwork()
+  const { type, chainId, ensRegistry, defaultEthNode } = getNetwork()
   return {
     name: type,
     chainId: chainId,
     ensAddress: ensRegistry,
+    _defaultProvider: defaultEthNode,
```
fabriziovigevani, line 124: Are we going to need this in the end? Thought that we saw that it expects a function

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "3box": "^1.19.1",
-    "@1hive/1hive-ui": "1.0.11",
+    "@1hive/1hive-ui": "^1.0.13",
```
fabriziovigevani, line 7: Let´s upgrade to the latest which includes the addition of arbiscan and polyscan explorers
```suggestion
    "@1hive/1hive-ui": "^1.0.14",
```

fabriziovigevani (approved):

PR #691, opened 2021-11-25
App: add garden loader

0xGabi:
Close https://github.com/1Hive/gardens/issues/437

https://user-images.githubusercontent.com/9082013/143332204-b6788a80-678b-4c43-b3a7-f2f95b1dff8d.mov


src/constants.js
```diff
@@ -18,6 +18,8 @@ export const PCT_BASE = bigNum(1)
 
 export const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
 
+export const HIVE_GARDEN_ADDRESS = '0x8ccbeab14b5ac4a431fffc39f4bec4089020a155'
```
fabriziovigevani, line 21: I think this should go for each network right? We won´t have one for Polygon but we do have for Rinkeby and xDai

src/components/Loader.js
```diff
@@ -1,25 +1,41 @@
...
+  left: 45%;
 `
 
 function Loader() {
+  const match = useRouteMatch('/garden/:daoId')
+
+  const is1HiveGarden = useMemo(() => {
+    if (match) {
+      const gardenAddress = match.params.daoId
+      return gardenAddress === HIVE_GARDEN_ADDRESS
```
fabriziovigevani, line 23: Let´s use `addressesEqual`

src/components/Loader.js
```diff
@@ -1,25 +1,41 @@
...
+import { HIVE_GARDEN_ADDRESS } from '@/constants'
 
 const Wrapper = styled.div`
   pointer-events: none;
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  height: 500px;
-  width: 100%;
+  position: fixed;
```
fabriziovigevani, line 12: We should use `absolute` instead.  `fixed` position is always relative to the viewport, doesn´t respect parent´s layout (e.g You can see that the loader for the list of gardens is off)

We would also need to add a parent div with a 100% height to the Garden Home and ProposalDetail which renders the Loader to properly center it in this case.

fabriziovigevani (commented):
💪

PR #693, opened 2021-11-25
Refactor connected garden

fabriziovigevani:
Creates a `ConnectedGarden` provider to separate logic between gardens and connected garden.

This prevents the Error Garden not found when selecting a Garden from the list. The useGarden should only be called when there´s a match on the route. 

It will also be useful when we implement aragonId support.


0xGabi (approved):
LGTM! Much clear this way 👍

PR #695, opened 2021-11-26
TrxHash for proposals

rperez89:
Screen Shot 2021-11-25 at 9 24 12 PM

### DIFF
--- a/src/components/Garden/DecisionDetail/DecisionDetail.js
+++ b/src/components/Garden/DecisionDetail/DecisionDetail.js
@@ -8,6 +8,7 @@ import {
   LoadingRing,
   Split,
   textStyle,
+  TransactionBadge,
   useLayout,
   useTheme,
 } from '@1hive/1hive-ui'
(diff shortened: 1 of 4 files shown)


src/components/Garden/DecisionDetail/DecisionDetail.js
```diff
@@ -145,6 +150,17 @@ function DecisionDetail({ proposal, actions }) {
...
                     {`Vote #${number}`}
                   </h1>
+                  <div
+                    css={`
+                      margin-top: ${1 * GU}px;
+                    `}
+                  >
+                    <TransactionBadge
+                      transaction={proposal.txHash}
+                      networkType={getNetworkType()}
```
fabriziovigevani, line 160: Let´s do the same as we do For the [`Step`](https://github.com/1Hive/honey-pot/blob/bd098f0df7c56360e413d8a3c2a367a588c5f918/src/components/Stepper/Step/Step.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L194) item and the `DeploymentStepItem` where we get the network before render and use the props.

so that we don´t import two functions.

src/components/Garden/ProposalDetail/ProposalDetail.js
```diff
@@ -177,6 +181,17 @@ function ProposalDetail({
...
                       {name}
                     </h1>
+                    <div
+                      css={`
+                        margin-top: ${1 * GU}px;
+                      `}
+                    >
+                      <TransactionBadge
+                        transaction={txHash}
+                        networkType={getNetworkType()}
```
fabriziovigevani, line 191: Same here

fabriziovigevani (commented):
Awesome, just some minor comment.

Also, curious why the yarn.lock wasn´t updated from the updated connect-gardens version installation

PR #696, opened 2021-11-26
Add help info about price oracle

fabriziovigevani:
closes 1Hive/gardens#321.

Addresses changes done in which where done against master and includes PR comments

### DIFF
--- a/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
+++ b/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
@@ -5,6 +5,7 @@ import {
   DropDown,
   Field,
   GU,
+  Help,
   Info,
   isAddress,
   Link,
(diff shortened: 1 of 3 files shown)


src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -392,6 +393,24 @@ function RequestedAmount({
             >
               <Checkbox checked={stable} onChange={onIsStableChange} />
               <span>Stable amount ({stableToken.symbol})</span>
+              {stable && (
```
0xGabi, line 396: I wonder if we should always show the Help tip.

src/components/Garden/ProposalDetail/ProposalDetail.js
```diff
@@ -622,7 +622,16 @@ const Amount = ({
                 {requestToken.symbol}
               </span>
               <Help hint="">
-                Converted to {requestToken.symbol} at time of execution
+                Converted to {requestToken.symbol} at time of execution. For
```
0xGabi, line 625: Probably we should add the same message to the `ProposalInfo`. What do you think?

0xGabi (changes requested):

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
fabriziovigevani: Let´s do it

src/components/Garden/ProposalDetail/ProposalDetail.js
fabriziovigevani: Ah yeas good catch

0xGabi (approved):

PR #698, opened 2021-11-26
Activity fixes

fabriziovigevani:
- Adds Allow manager action to garden action types and radspec. When allowing manager and opening the activity component it would error.
- Assets: Reuses `garden-action-types`


rperez89 (approved):
LGTM! nice one re using the assets from the action types

PR #699, opened 2021-11-26
Loader refactor

fabriziovigevani:
- Removes Layout height
- Adds GardenLoader to be used only by Garden Components. We can´t have a good solution for all loading screens when they expect different dimensions from the parent. 
- Moved hive addresses to networks file and moved is1HiveGarden logic to garden-utils


0xGabi (approved):
Nice!! Great to see how you solve it 🙌

PR #704, opened 2021-11-27
App: handle the right dex per network

0xGabi:


src/endpoints.js
```diff
@@ -26,5 +26,11 @@ export const IPFS_ENDPOINT = {
 export const GITHUB_ENDPOINT = 'https://github.com/'
 export const TWITTER_ENDPOINT = 'https://twitter.com/'
 
-export const getHoneyswapTradeTokenUrl = tokenAddress =>
-  `https://app.honeyswap.org/#/swap?outputCurrency=${tokenAddress}`
+export const getDexTradeTokenUrl = (chainId, tokenAddress) => {
+  switch (chainId) {
```
fabriziovigevani, line 30: Should we also add Uniswap for the Rinkeby instance?

fabriziovigevani (approved):
🙌

PR #706, opened 2021-11-27
App: several style fixes

0xGabi:
https://user-images.githubusercontent.com/9082013/143675177-fbf42d09-dde9-4220-a0ec-6dd784acce4a.mp4


src/components/BalanceModule.js
```diff
@@ -1,33 +1,25 @@
-import React, { useCallback } from 'react'
-import { useHistory } from 'react-router-dom'
+import React from 'react'
```
0xGabi, line 1: Remove the link to the deposit manager from the balance as it's not related. Especially when the user has a BYOT Garden.

src/components/Header/Header.js
```diff
@@ -203,16 +203,6 @@ function GardenNavItems({ garden }) {
           Wiki
         </Link>
       )}
-      <Link
```
0xGabi, line 206: Remove the documentation link because there is a link on the home screen and the header was already convoluted enough.

src/components/MainView.js
```diff
@@ -63,46 +63,44 @@ function MainView({ children }) {
             width: 100%;
           `}
         >
-          <Root.Provider>
+          <Root.Provider
+            css={`
```
0xGabi, line 67: Had to include this change back because otherwise, the `ScrrollView` doesn't work. Maybe there is another way of doing it, so we can also prevent the console warning. What do you think @fabriziovigevani?

src/components/Stepper/Stepper.js
```diff
@@ -163,17 +163,15 @@ function Stepper({ steps, onComplete, onCompleteActions }) {
...
-  // useEffect(() => {
-  //   let timeout
-  //   if (completed && !onCompleteActions) {
-  //     timeout = setTimeout(() => close(), 2000)
-  //   }
-  //   return () => {
-  //     clearTimeout(timeout)
-  //   }
-  // }, [close, completed, onCompleteActions])
+  useEffect(() => {
```
0xGabi, line 166: Now that we have notifications support. Add back again the autoclose handler on every modal.

src/components/Garden/Feed/WrapToken.js
```diff
@@ -11,22 +10,31 @@ import {
   useTheme,
 } from '@1hive/1hive-ui'
 
+import Carousel from '@/components/Carousel/Carousel'
```
0xGabi, line 13: Use the `Carousel` component to improve the wrap/unwrap/claim actions.

src/components/Garden/Feed/RightPanel.js
```diff
@@ -60,6 +60,26 @@ function RightPanel({
...
+          <WrapToken
+            onClaimRewards={onClaimRewards}
+            onUnwrapToken={onUnwrapToken}
+            onWrapToken={onWrapToken}
+          />
+        )}
+      </div>
+    )
+  }
+
```
0xGabi, line 82: @fabriziovigevani maybe you can help me figure out the right style so in case we are on `largeMode` we display the `Carousel` in a way that we see two buttons at a time. I think this is a good tradeoff because on the `largeMode` we have enough space. In other views, we should show a single button at a time.

src/components/Garden/Feed/WrapToken.js
fabriziovigevani:
```suggestion
import Carousel from '@components/Carousel/Carousel'
```

src/components/Garden/Feed/WrapToken.js
```diff
@@ -11,22 +10,31 @@ import {
...
 import unwrappedIcon from '@assets/unwrappedIcon.svg'
+import claimRewardsIcon from '@assets/rewardsWrapperIcon.svg'
 
-function WrapToken({ onClaimRewards, onUnwrapToken, onWrapToken }) {
+function WrapToken({
+  onClaimRewards,
+  onUnwrapToken,
+  onWrapToken,
+  itemWidth,
+  itemHeight,
```
fabriziovigevani, line 28: Seems we are not using them

src/components/Garden/Feed/WrapToken.js
```diff
@@ -39,74 +47,71 @@ function WrapToken({ onClaimRewards, onUnwrapToken, onWrapToken }) {
...
     </Box>
   )
 }
 
 function Token({ balance, loading, mode, onClick, token }) {
+  const theme = useTheme()
+
   const wrapMode = mode === 'wrap'
-  const icon = wrapMode ? unwrappedIcon : wrappedIcon
+  const claimMode = mode === 'claim'
```
fabriziovigevani, line 105: Whenever we have 3 possible values we can create a mapping on top based on the mode. 

```
const modeAttributes = {
  wrap: { icon: unwrappedIcon, button: { mode: 'strong', label: 'Wrap' } },
  unwrap: {
    icon: wrappedIcon,
    button: { mode: 'strong', label: 'Unwrap' },
    hint:
      'This amount can be used to vote on proposals. It can be unwrapped at any time.',
  },
  claim: {
    button: { mode: 'normal', label: 'Claim' },
    icon: claimRewardsIcon,
  },
}
```

In the example went ahead and added also the hint so we can use it as the condition for rendering the Help

src/components/Garden/Metrics.js
```diff
@@ -289,67 +296,47 @@ function TokenPrice({
...
+        {oracleText}
         {oracleMode && (
-          <Help hint="What is Oracle Price?">
-            This is the price used for funding proposals defined in stable
-            asset. Calling update price could return a reward.
-          </Help>
+          <span
+            css={`
+              padding-left: ${1 * GU}px;
+              display: inline-block;
```
fabriziovigevani, line 309: Let´s use margin instead, also the display attr seems unnecessary due to the parent div flex style
```suggestion
              margin-left: ${1 * GU}px;
```

src/components/Garden/Metrics.js
```diff
@@ -289,67 +296,47 @@ function TokenPrice({
...
-            `}
-          >
-            -
-          </span>
-        )}
+        {oraclePrice}
         {oracleMode && (
           <Button
-            size="small"
+            css={`
```
fabriziovigevani, line 330: We usually set the css prop as the last one

src/components/MainView.js
fabriziovigevani: Let´s keep it this way for now, we can address it later

src/components/Garden/Metrics.js
```diff
@@ -289,67 +296,47 @@ function TokenPrice({
...
-        >
-          {token.symbol} Price{' '}
-          {oracleMode && (
-            <span
-              css={`
-                ${textStyle('body4')}
-              `}
-            >
-              (Oracle)
-            </span>
```
fabriziovigevani, line 309: Let´s perhaps leave this smaller as it can cause overflow for the rest of the metrics. When we have the dedicated metrics , balances page this shouldn´t be a problem :D
Screen Shot 2021-11-28 at 22 44 31

fabriziovigevani (commented):
Awesome job 🚀 

Left some comments and suggestions

src/components/Garden/Feed/WrapToken.js
0xGabi: I didn't understand how to use it when passing as a prop 😅

PR #707, opened 2021-11-28
App(settings): add terminal executor

0xGabi:
https://user-images.githubusercontent.com/9082013/143724035-234fed6a-3640-492d-a76f-135f5078f464.mp4


src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -46,6 +50,24 @@ function EVMExecutor() {
   const [selectedApp, setSelectedApp] = useState(null)
   const [selectedFunction, setSelectedFunction] = useState(null)
   const [parameters, setParameters] = useState([])
+  const [code, setCode] = useState(
+    `# Available commands:
```
rperez89, line 54: can we have this text on the constant file?

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -216,9 +238,20 @@ function EVMExecutor() {
         { context: 'new decision' }
       )
     }
+    if (interactionType === TERMINAL_INDEX) {
+      intent = await evmcrispr.encode(
+        evmcl`${code}`,
+        [forwarderName],
+        // TODO: just for now that for some reason the radspec description on the card is not working, after fixed we can ask the user for enter some forum post related to why the decision is being created
```
rperez89, line 245: since we are duplicating this text maybe let's move to the top of the function

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -310,6 +343,44 @@ function EVMExecutor() {
...
+              setOptions={{
+                enableBasicAutocompletion: true,
+                enableLiveAutocompletion: true,
+                enableSnippets: true,
+                showLineNumbers: true,
+                tabSize: 2,
+              }}
+            />
+          </Box>
+          <Button
```
rperez89, line 371: we already have the button on line 401, let's just add another condition that when the interaction type is Terminal and the text inside the terminal is not null show it

rperez89 (changes requested):
nicely done gaby!

PR #710, opened 2021-11-28
Refactor wallet

fabriziovigevani:
PR which includes a minor refactor on Wallet provider.


src/hooks/useStaking.js
```diff
@@ -38,12 +38,12 @@ export function useStaking() {
     stakingFactoryAbi
   )
 
-  const stakingContract = useContract(
+  const stakingContract = useContractReadOnly(
```
fabriziovigevani, line 41: This is to prevent errors on cases where the connected garden network is different from the connected wallet network.

And since we have the Multimodal implementation, there´s no need to have the full contract with signer.

src/providers/Wallet.js
```diff
@@ -74,43 +111,9 @@ function WalletAugmented({ children }) {
...
-        setPreferredChain(wallet.chainId)
-        setPreferredNetwork(wallet.chainId)
-      }
-    }
-  }, [wallet.account, wallet.chainId, chainId])
-
-  /* Note that with this hook and the previous one we are manually reseting the connection on chain changed detected.
-  Here i noticed that is better to always depend on the data that comes from the use-wallet library instead  of just subscribing to window.ethereum.on('chainChanged', handleChainChanged) 
-  because for some reason the reconnect action is way faster and the windows event subscription was triggering many re renders */
-  useEffect(() => {
```
fabriziovigevani, line 98: Note that this effect is not needed as the `connect` function from useWallet already [resets](https://github.com/aragon/use-wallet/blob/d431fdec8e8f7b47867f74469bef900ff19423f4/src/index.tsx#L199) the connection.

src/providers/Wallet.js
```diff
@@ -124,40 +127,25 @@ function WalletAugmented({ children }) {
...
+  // This useEffect is needed because we don't have immediately available wallet.chainId right after connecting in the previous hook
+  // We just want to trigger this effect on wallet network change, so we´ll remove preferredNetwork from the useEffect dependencies
+  useEffect(() => {
+    if (wallet.account && preferredNetwork !== wallet.chainId) {
+      if (SUPPORTED_CHAINS.includes(wallet.chainId)) {
+        setPreferredChain(wallet.chainId)
+        setPreferredNetwork(wallet.chainId)
+      }
+    }
+  }, [wallet.account, wallet.chainId]) // eslint-disable-line
```
fabriziovigevani, line 142: Removed the preferredNetwork from the effect dependencies to ensure this is only triggered on network change or first render

src/providers/Wallet.js
```diff
@@ -165,4 +83,75 @@ function WalletProvider({ children }) {
...
 
+function useConnection() {
+  const wallet = useWallet()
+  const { account, connector } = wallet
+  /* We need  to pass down on the providers tree a preferred network in case that there is no network connnected
+  or the connected network is not supported in order to show some data and also to react to the network drop down selector changes */
+  const [preferredNetwork, setPreferredNetwork] = useState(getPreferredChain())
+  const [switchingNetworks, setSwitchingNetworks] = useState(false)
+
+  const connect = useCallback(
```
fabriziovigevani, line 94: Removed the `addEthereumChain` function call. The rationale being that we will only need to add a new chain if the user selects one of the chains in the dropdown of the garden list filters. In that case the user will be prompted to switch networks and the `switchNetwork` function already handles the network addition if it doesn´t exist.

0xGabi (approved):
LGTM!

src/components/Account/ScreenPromptingAction.js
```diff
@@ -89,7 +89,7 @@ const AccountModuleActionScreen = React.memo(function({ onCancel }) {
             color: ${theme.surfaceContentSecondary};
           `}
         >
-          {`Create the ${name} network in your provider and switch to it. You may be temporarily redirected to a new screen.`}
+          {`Switch to/ Create the ${name} network in your provider. You may be temporarily redirected to a new screen.`}
```
rperez89, line 92:
```suggestion
          {`Switch to create the ${name} network in your provider. You may be temporarily redirected to a new screen.`}
```

rperez89 (commented):
awesome refactor!

PR #718, opened 2021-12-01
Onboarding: Progress save

fabriziovigevani:
closes https://github.com/1Hive/gardens/issues/469

- Implemented a hook `useProgressSaver` which handles all the logic of storing the updated config as well as logic indicating wether the user wants to resume a saved progress.
- Implemented basic Box UI to ask the user if they would like to pick up from the last progress saved. 
 **Preview**

This is just a basic UI @fioreb if you can do your magic with this component 🙏  because im not a fan of the result or the texts either 😅 .
Screen Shot 2021-11-30 at 21 29 00

Things to consider: 

- When stringifying the config object so it can be stored locally, the logos blob File loses all data expect for the `path`(which indicates the file name). Because of this i implemented a asset recovery function `recoverAssets` which given the content and the name it recovers the original blob data.

### DIFF
--- a/src/utils/kit-utils.js
+++ b/src/utils/kit-utils.js
@@ -54,3 +54,23 @@ export const readFile = (reader, file) => {
       reader.readAsText(file)
   }
 }
+
+export function dataURLtoFile(dataurl, filename) {
+  var arr = dataurl.split(',')
+  var mime = arr[0].match(/:(.*?);/)[1]
+  var bstr = atob(arr[1])
+  var n = bstr.length
+  var u8arr = new Uint8Array(n)
+
+  while (n--) {
+    u8arr[n] = bstr.charCodeAt(n)
+  }
+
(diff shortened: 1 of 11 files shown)


src/utils/kit-utils.js
```diff
@@ -54,3 +54,23 @@ export const readFile = (reader, file) => {
       reader.readAsText(file)
   }
 }
+
+export function dataURLtoFile(dataurl, filename) {
+  var arr = dataurl.split(',')
+  var mime = arr[0].match(/:(.*?);/)[1]
+  var bstr = atob(arr[1])
+  var n = bstr.length
+  var u8arr = new Uint8Array(n)
```
0xGabi, line 63: Let's use `const` instead here.

0xGabi (changes requested):
Well done!

PR #728, opened 2021-12-03
Update terminal executor to use simple text input

0xGabi:


src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -307,31 +305,15 @@ function EVMExecutor({ evmcrispr }) {
...
-                enableLiveAutocompletion: true,
-                enableSnippets: true,
-                showLineNumbers: true,
-                tabSize: 2,
-              }}
-            />
-          </Box>
-        </>
+        <TextInput
+          onChange={setCode}
```
fabriziovigevani, line 309: We need to create a handler for this changebecause the argument will be the event object.

PR #731, opened 2021-12-07
Add typescript support

tmoutinho:


tmoutinho:
This was a fail attempt. the correct one was https://github.com/1Hive/gardens-ui/pull/738

PR #738, opened 2021-12-11
Ts support

tmoutinho:
## Topics

1. Added a prettier file for vscode to pick up on that, since we have a new config for eslint
2. Added support for typescript 
3. Css prop can be used in `tsx` too
4. The file `Home.tsx` has an example how to use styled components with typescript
5. From now on, when we develop new features or tasks we should try to migrate to typescript. 

## When refactoring a `js` file to `tsx` 

If you have a warning in @h1ve-ui lib, you will need to add its type in `/types/index.d.ts`

## Parcel

The main diference with Parcel is that requires zero configuration, which is good for all typescript projects, but not for legacy (`js` + `tsx`) projects like this. One other problem i had with this one is that css prop (which is all over the place) is not supported, even with emotion.

## Vitejs

This bundler is suppose to the fastest one, but i had a `jsx` blocker in all `js` files, and could not go through with this. Also the css prop support was not ideal. But with this experience I think I could manage integrating Vitejs now 😎

🤟⚡️


src/components/Home.tsx
```diff
@@ -13,8 +12,19 @@ import ConnectWalletScreens from './MultiModal/ConnectWallet/ConnectWalletScreen
...
+const DynamicSection = styled.div<{
+  marginTop: any
+}>`
+  margin-top: ${props => props.marginTop}px;
+  padding: 0 ${2 * GU}px;
+`
+
+const LoadingSection = styled.div`
+  margin: ${5 * GU}px 0;
+`
```
0xGabi, line 24: I think we already defined some styled-components this way and we decided to put them at the bottom after the main logic. What do you think?

src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
0xGabi, line 1: Why is this necessary?

src/hooks/usePromise.js
```diff
@@ -13,8 +13,6 @@ export default function usePromise(fn, memoParams, defaultValue) {
     return () => {
       cancelled = true
     }
-
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
0xGabi, line 17: I saw in a few places that we remove this rule without adding a global. Is it a safe assumption we are making for those cases?

src/components/Garden/Stake/StakingMovements.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-redeclare */
```
0xGabi, line 1: Why do we need to include this rule here?

tsconfig.json
```diff
@@ -0,0 +1,34 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
0xGabi, line 3: I wonder what is the tradeoff if we aim for a higher target, like `es6`.

0xGabi (commented):
Well done @tmoutinho. Thanks for working on this great improvement to the developer experience. The decision to do an incremental migration process was the right one 👍

Would you mind writing a brief summary of what you find out while trying to migrate to other bundlers (vite.js, parcel)? I think your experience could be a good resource to tap in a future time when we consider an upgrade on that front.

src/components/Home.tsx
tmoutinho: In my company we have add them to a separate file like header.styled.ts, and create a folder called Home and put the Header.tsx and the styled.ts inside it.

src/react-app-env.d.ts
tmoutinho: This was automatically added when I created the tsconfig. Basically it brings types declarations to react. I have created also one folder called types for those libs that have no type support like @h1ve-ui, in types/index.d.ts (it's not completed since I only need to add it for a couple of tsx files, but you get the idea). I also created one for styled components css prop in typescript.

src/hooks/usePromise.js
tmoutinho: This eslint rules is something it should not be global, since the correct way is to put all dependencies in the [], so if for some reason you want to override you must add this eslint disable line.
https://reactjs.org/docs/hooks-rules.html

tsconfig.json
tmoutinho: I added new target now 👌

src/components/Home.tsx
0xGabi: Interesting, we can ask the rest of the team how they feel with the different options and stick to one of them going forward 👍

0xGabi (approved):

src/components/Home.tsx
kamikazebr: Yah i'm used to make each style that way. Component and styled.ts separated, its get alot the folder but it get better refactor and less size by file.

kamikazebr: What I make different is use type or interface to props. So it can be:


```typescript
interface IDynamicSectionProps {
  marginTop: any
};

const DynamicSection = styled.div<IDynamicSectionProps>`
margin-top: ${props => props.marginTop}px;
padding: 0 ${2 * GU}px;
`

```

tmoutinho: @kamikazebr Yeah that's the same thing. But its the best way for sure 🔥

PR #739, opened 2021-12-13
Handle Decision delay execution

fabriziovigevani:
Screen Shot 2021-12-13 at 16 09 40
Screen Shot 2021-12-13 at 16 09 58

### DIFF
--- a/src/utils/vote-utils.js
+++ b/src/utils/vote-utils.js
@@ -69,7 +69,7 @@ export function getDelegatedVotingEndDate(vote) {
   const baseDelegatedVotingEndDate = vote.startDate + vote.delegatedVotingPeriod
 
   // If the vote was paused before the delegated voting period ended, we need to extend it
-  if (vote.pausedAt > 0 && vote.pausedAt > baseDelegatedVotingEndDate) {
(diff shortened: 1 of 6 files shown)


src/utils/vote-utils.js
```diff
@@ -65,6 +65,17 @@ export function hasVoteEnded(status, endDate, challengeEndDate) {
   )
 }
 
+export function getDelegatedVotingEndDate(vote) {
+  const baseDelegatedVotingEndDate = vote.startDate + vote.delegatedVotingPeriod
+
+  // If the vote was paused before the delegated voting period ended, we need to extend it
+  if (vote.pausedAt > 0 && vote.pausedAt > baseDelegatedVotingEndDate) {
```
0xGabi, line 72:
```suggestion
  if (vote.pausedAt > 0 && vote.pausedAt < baseDelegatedVotingEndDate) {
```

0xGabi: I think here we should use a `<` instead.

0xGabi (changes requested):
LGTM!

PR #741, opened 2021-12-15
Sidebar review

fabriziovigevani:


src/components/Garden/Agreement/Agreement.js
```diff
@@ -35,7 +35,7 @@ function Agreement() {
   }
 
   return (
-    <LayoutGutter>
+    <LayoutGutter collapseWhenSmall>
```
fabriziovigevani, line 38: Not related with the sidebar but from @PJColombo video realized we needed to adjust the padding

src/components/MainView.js
```diff
@@ -18,18 +18,16 @@ function MainView({ children }) {
...
+  const [showMobileSidebar, setShowMobileSidebar] = useState(false)
   const [createProposalModalVisible, setCreateProposalModalVisible] = useState(
     false
   )
-  let loadingGardenState = true
 
   const handleToggleSidebar = useCallback(() => {
-    setShowSidebar(prevShowSidebar => (!mobileMode ? true : !prevShowSidebar))
-  }, [mobileMode])
+    setShowMobileSidebar(prevShowMobileSidebar => !prevShowMobileSidebar)
```
fabriziovigevani, line 27: Updated this logic as we only toggle the mobile sidebar when we are in mobile mode, so don´t think the condition is necessary

src/components/MainView.js
```diff
@@ -85,9 +87,7 @@ function MainView({ children }) {
               flex-grow: 1;
               height: 100%;
               position: relative;
-              ${connectedGarden && !mobileMode
-                ? `margin-left: ${9 * GU}px;`
-                : ''}
+              ${hasSidebar && !mobileMode ? `margin-left: ${9 * GU}px;` : ''}
```
fabriziovigevani, line 90: This a more general fix as the margin was not being applied when we are in the profile page for example

src/components/Sidebars/InnerSidebars/InnerGardenNavigationSidebar.js
```diff
@@ -18,25 +14,19 @@ const InnerGardenNavigationSidebar = ({
   onToggle,
   onOpenCreateProposal,
 }) => {
-  const location = useLocation()
-  const { preferredNetwork } = useWallet()
-
-  const connectedGarden = useConnectedGarden()
-  const networkType = getNetworkType(
-    connectedGarden?.chainId || preferredNetwork
-  )
```
fabriziovigevani, line 27: Reverted this back as `buildGardenPath` already handles the network type in the regex, and we can assume that the current location will have a garden path since this is rendered

src/components/Sidebars/Items/BaseItem.js
```diff
@@ -49,27 +58,28 @@ const BaseItem = ({
...
         css={`
-          display: flex;
-          padding: ${1.5 * GU}px;
-          ${!active &&
-            `&:hover {
-            background: rgb(246, 246, 247);
-          }`}
+          display: block;
+          width: 100%;
+          color: inherit;
```
fabriziovigevani, line 67: Swapped positions the Link and div so that in the case of the inner garden nav items the selection  takes the whole row

src/components/Sidebars/MobileSidebar.js
```diff
@@ -73,20 +75,22 @@ const MobileSidebar = ({ show, onToggle, onOpenCreateProposal }) => {
...
-        <AnimatedSidebar
-          animationKey={key.concat(1)}
-          animationProps={props}
-          width={`${NAVIGATION_SIDEBAR_WIDTH}px`}
-          left={`${GARDENS_SIDEBAR_WIDTH}px`}
-          top={`${8 * GU}px`}
-          zIndex={3}
-        >
-          <InnerGardenNavigationSidebar
+        {connectedGarden && (
```
fabriziovigevani, line 78: We need to add this condition since we can be in the profile page as well

src/components/Sidebars/Items/BaseItem.js
PJColombo: Yeah, definitely better this way. I suppose I had a case of dyslexia here ^^!

PJColombo (approved):

PR #753, opened 2021-12-22
DX: add vercel env

0xGabi:
With these changes, we should be able to keep a simpler developer flow. Where `master` is the production deployment. And `staging` or `dev` is the branch we deploy to `gardens-staging`.


kamikazebr (approved):
It should work great!

PR #760, opened 2021-12-27
chore: update dependencies

### LINKED ISSUE #744: Update base dependencies of gardens-ui
Currently we are using out-dated version of the `react-app-rewired`, `customise-cra` which is a blocker for @tmoutinho for working on integrating [ceramic idx](https://github.com/ceramicstudio/self.id/issues/51).

I will be working on this task for updating the dependencies for gardens-ui.

Note: I gave it an estimate of 13 because I don't know the internals of the gardens-ui.

reznord:
### Description

As of right now, we are using outdated versions of bundler (webpack) and also not leveraging many of the optimisations like route based code-splitting using `React.lazy` etc.

Also, when @tmoutinho was working on integration [ceramic idx](#397), he bumped into an issue because of out-dated dependencies.

### What Changed?

This PR introduces a bunch of changes overall.

- Update dependencies
- Add support for route-based code splitting

#### Update dependencies

Earlier we were relying on react-scripts v2.x which was running on webpack v2.x which didn't have many optimisations what webpack offers today (one of those mainly being dynamic imports).

Along with this, we are updating all the other major dependencies of the project such as babel, react version to v17.x etc.

#### Add support for route-based code-splitting

Up until now, we are just bundling all the dependencies and the entire project into one single chunk which was bundling up to  around 1.8 MB in size. 

With this route-based code-splitting, we will be splitting the routes into smaller chunks and remove them from the main bundle. This main bundle is loaded only once and then the routes are fetched dynamically whenever the user plans on visiting that route.

### Issues effected

Closes #744 

Unblocks #742

### Note:

- Since we had to add the missing web3 compatibility packages in bundle, it increased by 200 KB extra in the common chunk. I will try to take a look into on how we can reduce that and try to chop the bundle size a bit more.
- Please delete `node_modules` and then run a fresh `yarn install` locally.


.eslintrc.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   extends: [
     'eslint:recommended',
-    'plugin:prettier/recommended',
+    // 'plugin:prettier/recommended',
```
reznord, line 6: Commenting it out temporarily because the imports sorting plugin which Felipe added was creating more issues with the bump in the version of the prettier packages.

config-overrides.js
```diff
@@ -1,18 +1,10 @@
...
-  override,
-  useEslintRc,
-  enableEslintTypescript,
-} = require('customize-cra')
+const { addWebpackAlias, useBabelRc, override } = require('customize-cra')
 
 module.exports = override(
   useBabelRc(),
-  useEslintRc(path.resolve(__dirname, '.eslintrc.js')),
-  enableEslintTypescript(),
```
reznord, line 15: We don't need to inject eslint separately because react-scripts migrated to `eslint-webpack-plugin` from v4.x on which reads the eslint config files from the root of the repository by default.

package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "eslint ./src",
     "sync-assets": "copy-aragon-ui-assets ./public",
     "bundlewatch": "bundlewatch",
-    "prettier-ai": "./node_modules/.bin/prettier --write 'src/**/*.{ts,tsx,js}' --loglevel=error --ignore-path .prettier-imports-ignore"
```
reznord, line 11: we don't need to point to `node_modules` here separately because by default any of the scripts will always point to node_modules itself.

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -1,6 +1,6 @@
+/* eslint-disable no-unexpected-multiline */
```
reznord, line 1: Had to disable this eslint rule temporarily because it was giving errors locally when running a build.

src/components/SingleDatePicker/DatePicker.js
```diff
@@ -1,9 +1,10 @@
+/* eslint-disable no-unexpected-multiline */
```
reznord, line 1: Had to disable this eslint rule because it was giving errors locally when running a build.

src/routes/Routes.js
```diff
@@ -1,17 +1,28 @@
...
 import { Redirect, Route, Switch } from 'react-router-dom'
-import Garden from '@components/Garden'
-import Home from '@components/Home'
-import Profile from '@components/Profile/Profile'
+
+import { LoadingRing, Main } from '@1hive/1hive-ui'
+
+const Home = lazy(() => import('@components/Home'))
+const Garden = lazy(() => import('@components/Garden'))
+const Profile = lazy(() => import('@components/Profile/Profile'))
```
reznord, line 8: This is the suggested way to code-split routes in SPAs created using create-react-app or any project which internally uses react-scripts.

Reference link: https://reactjs.org/docs/code-splitting.html#route-based-code-splitting

src/types/styled-components.d.ts
```diff
@@ -1,9 +1,10 @@
-import { CSSProp } from "styled-components"
+import { CSSProp } from 'styled-components'
 
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
```
reznord, line 3: I had to disable this eslint rule since by default typescript-eslint doesn't allow to declare empty interfaces.

tsconfig.json
```diff
@@ -17,11 +17,12 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
reznord, line 20: for using the latest JSX features, we have to start using `react-jsx` or typescript won't be able to identify the jsx v2 syntax which is supported in react already.

reznord:
we need to clear the caches (`build` and `node_modules`) on vercel and the deploy will work without any issues.

reznord:
we need to clear the caches (`build` and `node_modules`) on vercel and the deploy will work without any issues.

cc @0xGabi @fabriziovigevani

package.json
kamikazebr: In my env it was getting the global one, idk why, but i'll remove it globally and should solve :D

tmoutinho (approved):
Also #742 is unblocked now.

kamikazebr (approved):
Works like a charm! 
PS: Need remove node_modules before yarn.

0xGabi (changes requested):
Amazing work @reznord the improvements on the app speed are already noticeable. Looking forward to what we can achieve once we migrate to `next.js` and keep reducing the bundle size 🙌

I notice anyways there is a small issue with the new changes that introduced suspense. Have a look at the following clip:

https://user-images.githubusercontent.com/9082013/147610139-095b6183-30a7-4346-b964-6e8605ebef2f.mp4

0xGabi (approved):

PR #764, opened 2021-12-28
Create Proposal Modal: Add go to proposal button 

rperez89:
Screen Shot 2021-12-28 at 11 43 26 AM


src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -229,12 +231,12 @@ function EVMExecutor({ evmcrispr }) {
     code,
   ])
 
-  const handleOnContractAddressChange = useCallback(event => {
+  const handleOnContractAddressChange = useCallback((event) => {
```
tmoutinho, line 234: Just a comment, that you dont need to use useCallback for a simple function like this. In my opinion since is a simple setState, you can do it in the onChange itself, `onChange={(e) => setExternalContractAddress(e.target.value)}`

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -160,7 +162,7 @@ function EVMExecutor({ evmcrispr }) {
 
   const handleOnChangeParameters = useCallback((index, event) => {
     const newValue = event.target.value
-    setParameters(prevState => {
+    setParameters((prevState) => {
```
tmoutinho, line 165: simply do it `=> [...prevState, [index]: newValue]`

tmoutinho (approved):
Approved with small suggestions :) Nice work.

src/components/Garden/Preferences/EVMExecutor.js
tmoutinho: Also in this file we are using 9 useState, we should consider using a reducer to simplify.

src/components/Garden/Preferences/EVMExecutor.js
kamikazebr: useCallback is better performance against anonymous function on onChange.
Plus its better to future changes too.

just my 2 honeys on that.

tmoutinho: You don't need to memoize a setState

src/components/Garden/Preferences/EVMExecutor.js
```diff
@@ -160,7 +162,7 @@ function EVMExecutor({ evmcrispr }) {
...
       const newArray = [...prevState]
       newArray[index] = newValue
       return newArray
```
kamikazebr, line 168:
```suggestion
      return  [...prevState, [index]: newValue]
```

tmoutinho: hehe @kamikazebr look up.

kamikazebr: Yah i just put it more clear where and how modify your suggestion, was it you suggested right?

tmoutinho: I didn't know that we could suggest the change on the code, nice 😎

src/components/Garden/Preferences/EVMExecutor.js
kamikazebr: I agree with you @tmoutinho - but with that reasons https://dmitripavlutin.com/dont-overuse-react-usecallback/

src/components/SingleDatePicker/DatePicker.js
```diff
@@ -1,9 +1,14 @@
+/* eslint-disable no-unexpected-multiline */
```
kamikazebr, line 1: Its just needed to compile. The correct way should be remove it and change the code, but it should be made on new PR.

src/components/Garden/Preferences/EVMExecutor.js
tmoutinho: We are using useCallback every time, and it's not needed, thats my point

src/components/SingleDatePicker/DatePicker.js
tmoutinho: This is removed in Anup's branch.

src/components/Garden/ModalFlows/ModalFlowBase.js
```diff
@@ -71,6 +77,8 @@ function ModalFlowBase({
                   // TODO: Provide a better user experience than waiting on all transactions
                   await tx.wait()
 
+                  setAccountSetting('lastTxHash', account, chainId, tx.hash)
```
kamikazebr, line 80: Maybe that string 'lastTxHash' put on some Enum inside local-settings?

Or convert that file to typescript and use 'lastTxHash' on type. 
What yours think about @rperez89 and @tmoutinho ?

src/components/Garden/Preferences/EVMExecutor.js
kamikazebr: Yah, if you want make try, i can cancel mine. Why if he accept it put my name like co-author, should be your name there.

src/components/Garden/ModalFlows/ModalFlowBase.js
tmoutinho: IMO we should try to migrate this files to typescript, but I understand that @rperez89 has no experience on that. Perhaps after this is merged I can try do migrate this files, as an example how we should do it. WDYT?

src/components/SingleDatePicker/DatePicker.js
kamikazebr: Good! Here still need.

tmoutinho: yeah, but let's keep it here, to not through conflict with anup PR.

src/components/Garden/Preferences/EVMExecutor.js
tmoutinho: Naa mate, go for it, we are a team, it doesn't matter you does it ;)

src/components/Garden/ModalFlows/ModalFlowBase.js
kamikazebr: Its look a good idea @tmoutinho. I'll need change the confetti thing to TS too, so we will have 2 examples on that.

kamikazebr (approved):
Tested locally! Works like a charm!

PR #770, opened 2022-01-03
Remove garden-connect extends types and use the default one

tmoutinho:
This fixes the error of garden-connect

Captura de ecrã 2022-01-03, às 22 22 40


0xGabi (approved):
LGTM!

PR #771, opened 2022-01-04
Migrate task #764 to typescript

tmoutinho:
The task was to migrate the task of @rperez89 to typescript to show how it should be done, from now on.

### DIFF
--- a/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
+++ b/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
@@ -43,7 +43,7 @@ const DEFAULT_FORM_DATA = {
 
 const PROPOSAL_TYPES = ['Suggestion', 'Funding']
 
-const AddProposalPanel = React.memo(({ setProposalData }) => {
+const AddProposalPanel = ({ setProposalData }) => {
   const { next } = useMultiModal()
   const { commonPool, config } = useGardenState()
   const {
@@ -64,8 +64,8 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
   const forumRegex = new RegExp(connectedGarden.forumURL)
 
   const handleAmountEditMode = useCallback(
-    editMode => {
-      setFormData(formData => {
+    (editMode) => {
+      setFormData((formData) => {
         const { amount } = formData
 
         const newValue = amount.valueBN.gte(0)
@@ -94,20 +94,20 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
     [stakeToken]
   )
 
-  const handleIsStableChange = useCallback(checked => {
-    setFormData(formData => ({
+  const handleIsStableChange = useCallback((checked) => {
+    setFormData((formData) => ({
       ...formData,
       amount: { ...formData.amount, stable: checked },
     }))
   }, [])
 
-  const handleTitleChange = useCallback(event => {
+  const handleTitleChange = useCallback((event) => {
     const updatedTitle = event.target.value
-    setFormData(formData => ({ ...formData, title: updatedTitle }))
+    setFormData((formData) => ({ ...formData, title: updatedTitle }))
   }, [])
 
   const handleAmountChange = useCallback(
-    event => {
+    (event) => {
       const updatedAmount = event.target.value
 
       const newAmountBN = new BigNumber(
@@ -116,7 +116,7 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
           : toDecimals(updatedAmount, stakeToken.decimals)
       )
 
-      setFormData(formData => ({
+      setFormData((formData) => ({
         ...formData,
         amount: {
           ...formData.amount,
@@ -128,26 +128,29 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
     [stakeToken.decimals]
   )
 
-  const handleProposalTypeChange = useCallback(selected => {
-    setFormData(formData => ({
+  const handleProposalTypeChange = useCallback((selected) => {
+    setFormData((formData) => ({
       ...formData,
       proposalType: selected,
     }))
   }, [])
 
-  const handleBeneficiaryChange = useCallback(event => {
+  const handleBeneficiaryChange = useCallback((event) => {
     const updatedBeneficiary = event.target.value
 
-    setFormData(formData => ({ ...formData, beneficiary: updatedBeneficiary }))
+    setFormData((formData) => ({
+      ...formData,
+      beneficiary: updatedBeneficiary,
+    }))
   }, [])
 
-  const handleLinkChange = useCallback(event => {
+  const handleLinkChange = useCallback((event) => {
     const updatedLink = event.target.value
-    setFormData(formData => ({ ...formData, link: updatedLink }))
+    setFormData((formData) => ({ ...formData, link: updatedLink }))
   }, [])
 
   const handleOnContinue = useCallback(
-    event => {
+    (event) => {
       event.preventDefault()
 
       setProposalData(formData)
@@ -323,7 +326,7 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
       )}
     </form>
   )
-})
+}
 
 function RequestedAmount({
   amount,
--- a/src/components/Garden/ModalFlows/CreateProposalScreens/CreateProposalScreens.tsx
+++ b/src/components/Garden/ModalFlows/CreateProposalScreens/CreateProposalScreens.tsx
@@ -24,7 +24,7 @@ import CreateProposalRequirements from './CreateProposalRequirements'
 
 function GoToProposal() {
   const history = useHistory()
-  const [proposalId, setProposalId] = useState()
+  const [proposalId, setProposalId] = useState<string>()
   const { account, chainId, ethers } = useWallet()
   const mounted = useMounted()
   const txHash = getAccountSetting('lastTxHash', account, chainId)
@@ -33,7 +33,7 @@ function GoToProposal() {
     async function getProposalId() {
       const id = await extractProposalId(ethers, txHash, 'conviction')
 
-      if (mounted) {
+      if (mounted()) {
         setProposalId(fromDecimals(id.toString(), 18))
       }
     }
@@ -55,7 +55,7 @@ function GoToProposal() {
   )
 }
 
-function CreateProposalScreens({ onComplete }) {
+function CreateProposalScreens({ onComplete }: { onComplete: () => void }) {
   const [loading, setLoading] = useState(true)
   const [transactions, setTransactions] = useState([])
   const { account } = useWallet()
@@ -63,7 +63,7 @@ function CreateProposalScreens({ onComplete }) {
   const { stakeManagement, loading: stakingLoading } = useStakingState()
   const { convictionActions } = useActions()
 
-  const proposalData = useRef()
+  const proposalData = useRef<any>()
 
   useEffect(() => {
     setLoading(true)
@@ -75,9 +75,9 @@ function CreateProposalScreens({ onComplete }) {
     setLoading(agreementLoading || stakingLoading)
   }, [agreementLoading, stakingLoading])
 
-  const handleSetProposalData = useCallback((data) => {
+  const handleSetProposalData = (data: any) => {
     proposalData.current = data
-  }, [])
+  }
 
   const onCompleteMiddleware = useCallback(() => {
     throwConfetti({
@@ -115,7 +115,7 @@ function CreateProposalScreens({ onComplete }) {
         }
       }
 
-      await convictionActions[fn](params, (intent) => {
+      await convictionActions[fn](params, (intent: any) => {
         setTransactions(intent)
         onComplete()
       })
--- a/src/components/Garden/ModalFlows/ModalFlowBase.tsx
+++ b/src/components/Garden/ModalFlows/ModalFlowBase.tsx
@@ -1,6 +1,4 @@
 import React, { useEffect, useMemo } from 'react'
-
-import PropTypes from 'prop-types'
 import { keyframes } from 'styled-components'
 
 import { GU, useTheme } from '@1hive/1hive-ui'
@@ -14,13 +12,18 @@ import { useActivity } from '@providers/ActivityProvider'
 import { useWallet } from '@providers/Wallet'
 
 import { setAccountSetting } from '@/local-settings'
+import { TransactionType } from '@/hooks/useActions'
 
-const indexNumber = {
-  0: 'First',
-  1: 'Second',
-  2: 'Third',
-  3: 'Fourth',
-  4: 'Fifth',
+const indexNumber = ['First', 'Second', 'Third', 'Fourth', 'Fifth']
+
+type ModalFlowBaseType = {
+  frontLoad: boolean
+  loading: boolean
+  screens: Array<any>
+  transactions: Array<any>
+  transactionTitle: string
+  onComplete: () => void
+  onCompleteActions: React.ReactNode
 }
 
 function ModalFlowBase({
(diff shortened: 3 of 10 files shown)


src/components/Garden/ModalFlows/CreateProposalScreens/CreateProposalScreens.tsx
```diff
@@ -33,7 +33,7 @@ function GoToProposal() {
     async function getProposalId() {
       const id = await extractProposalId(ethers, txHash, 'conviction')
 
-      if (mounted) {
+      if (mounted()) {
```
kamikazebr, line 36: Why removed the parenteses? Its look still a method defined on the hooks.

Btw, its possible rename the var to isMounted instead? if it change to boolean?

tmoutinho: I added the parênteses since it’s a function, we could just use useEffect, but didn’t want to change more the code. 

I don’t know how it’s working right now 🤷🏼‍♂️

src/hooks/useActions.ts
```diff
@@ -29,7 +30,9 @@ const SIGN_GAS_LIMIT = 100000
 const STAKE_GAS_LIMIT = 250000
 const WRAP_GAS_LIMIT = 1000000
 
-export default function useActions() {
+export default function useActions(): {
+  [x: string]: any
```
rperez89, line 34: why is this needed since is not receiving any parm?

src/hooks/useActions.ts
```diff
@@ -739,10 +742,18 @@ export default function useActions() {
 }
 
 async function sendIntent(
-  app,
-  fn,
-  params,
-  { ethers, from, gasLimit = GAS_LIMIT }
+  app: any,
```
rperez89, line 745: could we define the types outside the function definition to avoid those huge multiline function signatures?

src/components/Garden/ModalFlows/CreateProposalScreens/CreateProposalScreens.tsx
kamikazebr: Oh! i read inverse?. Make sense

src/providers/Staking.tsx
```diff
@@ -1,12 +1,19 @@
...
+}
+
+const StakingContext = React.createContext<State>({
+  stakeManagement: null,
+  stakeActions: null,
+  loading: false,
 })
 
-function StakingProvider({ children }) {
+function StakingProvider({ children }: { children: React.ReactNode }) {
```
rperez89, line 16: same here could we keep a coherence with how we define the function signatures with the types? so we all try to do the same way? for example i am defining everything outside the function line

rperez89 (commented):
Awesome tiago thxs for the intro, i have the global preference almost done from before going to vacations and think i am in the right path :)

src/components/Garden/ModalFlows/ModalFlowBase.tsx
```diff
@@ -14,13 +12,18 @@ import { useActivity } from '@providers/ActivityProvider'
...
-  0: 'First',
-  1: 'Second',
-  2: 'Third',
-  3: 'Fourth',
-  4: 'Fifth',
+import { TransactionType } from '@/hooks/constants'
+
+const indexNumber = ['First', 'Second', 'Third', 'Fourth', 'Fifth']
+
+type ModalFlowBaseType = {
```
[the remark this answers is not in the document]
kamikazebr: Yah I prefer interface over types too. Interface can be extended and merged if needed. 
More OOP to reuse, refactor and goodie stuffs. 
And Types used to union, interssection stuffs and more closed types.

Exist some excellent article on internet talking a lot about

One reference: https://blog.logrocket.com/types-vs-interfaces-in-typescript/

I think for now its do not do warm keep in Type, but we should know well the difference to know when need to change it :D

src/hooks/useActions.ts
kamikazebr: Its typed return of function, note its after `():`

Its type its kinda generic object, dynamic properties string based, the better way should be, what its exactly returns.

src/hooks/useActions.ts
kamikazebr: Yes its recommended. Plus the `fn: any,` its kinda enums of string, right?

You can define the type like 
```typescript
fn: 'executeAdjustment' | 'executeVote' | 'resolve'
```
and the IDE will pop-up it while write the method to choose.

Yet better way to write that its export that chunk to type so

```typescript
type AppFunctions = 'executeAdjustment' | 'executeVote' | 'resolve'
fn: AppFunctions

```
Amazing huh?

src/utils/proposal-utils.ts
```diff
@@ -30,30 +35,37 @@ export function hasProposalEnded(status, challengeEndDate) {
...
-export function getProposalStatusData(proposal) {
-  const statusData = {}
+export function getProposalStatusData(proposal: {
+  status: string
+  challengeEndDate: Date | any
+}) {
+  let statusData = {}
   if (proposal.status === PROPOSAL_STATUS_EXECUTED_STRING) {
-    statusData.executed = true
+    statusData = { ...statusData, executed: true }
```
kamikazebr, line 44: I think those expansions its not necessary.

typing statusData should be sufficient to keep the old code

yarn.lock
```diff
@@ -20144,9 +20144,9 @@ ursa-optional@~0.10.0:
     bindings "^1.5.0"
     nan "^2.14.2"
 
-"use-inside@git+https://github.com/aragon/use-inside.git#16f321e499d8f89e9b0fb0999e2d21db2b119bbf":
+"use-inside@https://github.com/aragon/use-inside#16f321e499d8f89e9b0fb0999e2d21db2b119bbf":
```
kamikazebr, line 20147: curious: why it automatic changed? Updated the yarn last version? I have 1.22.5

src/hooks/useActions.ts
kamikazebr: Looks like already exist some type we can use on return that, its right?

https://github.com/1Hive/gardens-ui/pull/771/files?diff=split&w=0#diff-04581b471ddd0404521d07d207d7ceb2529e3c4d9d73550d8bf8698ed09f51b9R234

src/components/Garden/ModalFlows/ModalFlowBase.tsx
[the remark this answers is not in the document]
tmoutinho: It's a matter of opinion and choosing. I like better to use Types since i use a lot enum's and that way we have types all the way. Both interface and type can use extends and union. So its comes to your choosing and opinion. Anyone is correct.

src/components/Garden/ModalFlows/ModalFlowBase.tsx
```diff
@@ -135,17 +143,16 @@ function ModalFlowBase({
...
   const { next } = useMultiModal()
 
   useEffect(() => {
-    let timeout
+    let timeout: number
 
     if (!loading) {
       // Provide a minimum appearance duration to avoid visual confusion on very fast requests
-      timeout = setTimeout(() => {
+      timeout = window.setTimeout(() => {
```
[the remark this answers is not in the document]
tmoutinho: Because we migrate this file to tsx, and since timeout is a number, we cannot assign a timeout variable to a setTimeout. The ts error is the following. `Type 'Timeout' is not assignable to type 'number' `

src/hooks/useActions.ts
tmoutinho: Yes exactly what Felipe said. I didn't know the correct type so I added any, but we can improve in the future. I will add that as Felipe suggested.

src/providers/Staking.tsx
tmoutinho: Yes, for coherence we should do that, although when its a simple children i dont often do that. But I will do it in Gardens :)

src/utils/proposal-utils.ts
tmoutinho: I am doing this for the sake of `statusData` not losing any data.
.

yarn.lock
tmoutinho: I dont know either, I didn't change anything. Just removed `node_modules` and run `yarn`

src/hooks/constants.ts
```diff
@@ -229,4 +231,97 @@ type ProposalType = {
   yeas: any
 }
 
-export type { ConfigType, FiltersType, TokenType, AppType, ProposalType }
+type ActionsType = {
```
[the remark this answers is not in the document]
tmoutinho: I created that, and forgot to replace 😄

src/hooks/useActions.ts
tmoutinho: I created that ActionsType but forgot to add to useActions. Already fixed.

src/utils/proposal-utils.ts
kamikazebr: How it will lose data?

yarn.lock
kamikazebr: Yah i noticied it two times in local, well i think its not issue, i'll check it later

src/components/Garden/ModalFlows/ModalFlowBase.tsx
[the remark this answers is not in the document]
kamikazebr: `Nodejs.setTimeout()
`
image

Exist 2 types of setTimeout, one from NodeJS and another from window, browser side.
So the types changes a bit.

`window.setTimeout()
`
image

src/utils/proposal-utils.ts
tmoutinho: Whats is the problem of returning a spread object?

tmoutinho: Whats is the problem of returning a spread object?

kamikazebr: Want know where it will lose data.

Already told, its not necessary. its verbose, increase code, hard legibility, for what? Nothing.

tmoutinho: This function itself is all that.

tmoutinho: I have changed the function

src/components/Garden/ModalFlows/ModalFlowBase.tsx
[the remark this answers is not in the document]
tmoutinho: The explanation of Felipe is correct, although in this case we use only client side, and the return is a positive integer value, that can be used by clearTimeout() to cancel the timer.

src/utils/proposal-utils.ts
kamikazebr: I think its improved a lot

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -43,7 +43,7 @@ const DEFAULT_FORM_DATA = {
 
 const PROPOSAL_TYPES = ['Suggestion', 'Funding']
 
-const AddProposalPanel = React.memo(({ setProposalData }) => {
+const AddProposalPanel = ({ setProposalData }) => {
```
kamikazebr, line 46: React.memo was moved to the end on export.

src/components/Garden/ModalFlows/CreateProposalScreens/CreateProposalScreens.tsx
```diff
@@ -93,32 +93,35 @@ function CreateProposalScreens({ onComplete }) {
...
-        }
+        await convictionActions.newProposal(
+          {
+            title,
+            link,
+            amount: convertedAmount,
+            stableRequestAmount,
+            beneficiary,
+          },
+          (intent: any) => {
```
kamikazebr, line 119: Looks great and more safe that way, what i'll suggest is encapsulate the callback

```
const callbackSomethingNameHere = (intent: any) => {
  setTransactions(intent)
            onComplete()
          }
```

So it will keep equal the previous one and avoid update one and forget the other.

tmoutinho: Yep makes sense. will update that.

tmoutinho: Just updated!

kamikazebr (approved):
LGTM!

src/components/Garden/ModalFlows/ModalFlowBase.tsx
```diff
@@ -14,32 +12,50 @@ import { useActivity } from '@providers/ActivityProvider'
 import { useWallet } from '@providers/Wallet'
 
 import { setAccountSetting } from '@/local-settings'
+import { TransactionType } from '@/hooks/constants'
+
+const indexNumber = ['First', 'Second', 'Third', 'Fourth', 'Fifth']
```
rperez89, line 17: since is a constant lets take the opportunity to use uppercase on the const name

rperez89, line 17:
```suggestion
const INDEX_NUMBER = ['First', 'Second', 'Third', 'Fourth', 'Fifth']
```
```suggestion
const indexNumber = ['First', 'Second', 'Third', 'Fourth', 'Fifth']
```

src/components/MultiModal/MultiModalProvider.tsx
```diff
@@ -1,16 +1,44 @@
...
+  getScreen: any
+  next: any
+  prev: any
+  step: any
+}
+
+const MultiModalContext = React.createContext<State>({
+  currentScreen: null,
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  close: () => {},
```
rperez89, line 20: can't we define as  `() => void` ?

src/hooks/useActions.ts
```diff
@@ -738,11 +739,17 @@ export default function useActions() {
   )
 }
 
+type AppFunctions = 'executeAdjustment' | 'executeVote' | 'resolve'
```
rperez89, line 742: i wouldn't define which app functions for this one because is something really generic, at some point we might use it a lot to do different contract calls and we don't want to add a bunch of funcion names to that type

src/hooks/useActions.ts
```diff
@@ -738,11 +739,17 @@ export default function useActions() {
...
+type OptionsIntent = {
```
rperez89, line 743:
```suggestion
type OptionsIntent = {
```
```suggestion
type IntentArguments = {
```

src/hooks/useActions.ts
```diff
@@ -754,13 +761,24 @@ async function sendIntent(
   }
 }
 
-function imposeGasLimit(intent, gasLimit) {
+function imposeGasLimit(intent: IntentType, gasLimit: number) {
```
rperez89, line 764: here the same, i would't define a type for an intent, it can be anything

rperez89 (changes requested):
awesome! Left some  final comments, also i am a bit worried because touched the useAction which means that touch all the smart contract interactions which means that the whole app need to be tested

src/hooks/useActions.ts
tmoutinho: Will fix it.

src/hooks/useActions.ts
tmoutinho: Sure. I will fix it now

src/hooks/useActions.ts
tmoutinho: Will change it to string. Thats ok?

src/components/MultiModal/MultiModalProvider.tsx
tmoutinho: This is not a type. It's an empty function :) The `() => void` is in the definition of the type.

src/components/Garden/ModalFlows/ModalFlowBase.tsx
tmoutinho: Fixed already :)

src/hooks/useActions.ts
rperez89: yeah

rperez89:
Let's resolve the conflicts and merge this one

PR #772, opened 2022-01-05
Disable sourcemap generation on development and on production

### LINKED ISSUE #769: Fix source map warnings
We should try and fix those warnings on build/start

Captura de ecrã 2022-01-03, às 18 30 46

reznord:
Right now, this is a common situation with the latest version of the create-react-app since the way source-map-loader is loaded by webpack (ref: https://github.com/mswjs/msw/issues/1030#issuecomment-995760580).

So, for the time being it is better to disable source-maps generation in gardens-ui.

Closes #769 

Note: Please remove the build folder `rm -rf build` and then run the `yarn start` or `yarn build`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "yarn run sync-assets && react-app-rewired start",
-    "build": "yarn run sync-assets && react-app-rewired build",
+    "start": "yarn run sync-assets && GENERATE_SOURCEMAP=false react-app-rewired start",
+    "build": "yarn run sync-assets && GENERATE_SOURCEMAP=false react-app-rewired build",
     "lint": "eslint ./src",
     "sync-assets": "copy-aragon-ui-assets ./public",
     "bundlewatch": "bundlewatch",
@@ -94,6 +94,7 @@
     "eslint-plugin-react-hooks": "^4.3.0",
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "eslint-plugin-standard": "^5.0.0",
+    "immer": "^9.0.7",
     "prettier": "^2.5.1",
     "react-app-polyfill": "^3.0.0",
     "react-app-rewire-provide-plugin": "^1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -94,13 +94,15 @@
     "eslint-plugin-react-hooks": "^4.3.0",
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "eslint-plugin-standard": "^5.0.0",
+    "immer": "^9.0.7",
```
reznord, line 97: need to remove this since we are disabling the sourcemaps.

package.json
```diff
@@ -94,13 +94,15 @@
...
     "prettier": "^2.5.1",
     "react-app-polyfill": "^3.0.0",
     "react-app-rewire-provide-plugin": "^1.0.0",
     "react-app-rewired": "^2.1.9",
     "react-dev-utils": "^12.0.0",
     "react-scripts": "^5.0.0",
     "rimraf": "^2.6.2",
+    "source-map-loader": "https://github.com/Volune/source-map-loader.git#fixes",
```
reznord, line 105: need to remove this since we are disabling the sourcemaps.

tmoutinho (approved):

kamikazebr (approved):
Building and Starting without warnings! Nice job!

yarn.lock
```diff
@@ -20144,9 +20144,9 @@ ursa-optional@~0.10.0:
     bindings "^1.5.0"
     nan "^2.14.2"
 
-"use-inside@git+https://github.com/aragon/use-inside.git#16f321e499d8f89e9b0fb0999e2d21db2b119bbf":
+"use-inside@https://github.com/aragon/use-inside#16f321e499d8f89e9b0fb0999e2d21db2b119bbf":
```
rperez89, line 20147: wondering why the lock changed with no new package installed?

rperez89 (commented):
LGTM just one question about the lock file

yarn.lock
tmoutinho: I don’t really know why this was updated. I always run yarn and delete node_modules.

reznord: it is because in the latest version of yarn, they started supporting the installing of npm modules from github repository directly using `https://....` URLs. Earlier they used to support the old way `git+ssh://....` or `git+https://...`.

This should have been fixed in the 1hive-ui lib itself but then it wasn't for some reason. Will fix that when I am working on the updating of the migration to peerDeps (reducing the size of the library.

PR #774, opened 2022-01-05
Ethereum connectors: remove mumbai

rperez89:
For some reason we were supporting mumbai as one of the allowed chains passed to useWallet now that we are testing the Maroon5 specific gardens we added the mumbai network to metamask and noticed that breaks the app because there is not data  defined for the app on the `networks` file under mumbai attr.

So since the M5 gardens instance is totally apart from our flow and has it own branch i am just removing it.

### DIFF
--- a/src/ethereum-providers/connectors.ts
+++ b/src/ethereum-providers/connectors.ts
@@ -10,13 +10,13 @@ export const CONNECTORS = [
   {
     id: 'injected',
     properties: {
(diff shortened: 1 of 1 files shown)


reznord (approved):
LGTM 👍🏻

tmoutinho (approved):
✌🏻

kamikazebr (approved):

0xGabi (approved):

PR #777, opened 2022-01-06
Fix re-rendering of list

tmoutinho:
This error was introduced while migrating the hooks folder.

### DIFF
--- a/src/hooks/useSubscriptions.ts
+++ b/src/hooks/useSubscriptions.ts
@@ -15,7 +15,7 @@ type Garden = {
 export function useConfigSubscription(garden: Garden | null) {
   const [config, setConfig] = useState(null)
 
-  const rawConfigRef = useRef<any>(null)
+  const rawConfigRef = useRef<unknown>(null)
   const configSubscription = useRef<any>(null)
 
   const onConfigHandler = useCallback((err, config) => {
@@ -51,8 +51,7 @@ export function useConfigSubscription(garden: Garden | null) {
 
 export function useProposalsSubscription(filters: FiltersType) {
   const { config, connector } = useGardenState()
-  const [proposals, setProposals] = useState<any>([])
-
+  const [proposals, setProposals] = useState<unknown>([])
   const proposalsSubscription = useRef<any>(null)
 
   const onProposalsHandler = useCallback(
@@ -111,7 +110,7 @@ export function useProposalSubscription(
   const [proposal, setProposal] = useState(null)
   const [loading, setLoading] = useState(true)
 
-  const rawProposalRef = useRef<any>(null)
+  const rawProposalRef = useRef<unknown>(null)
   const proposalSubscription = useRef<any>(null)
 
   const onProposalHandler = useCallback(
@@ -158,7 +157,7 @@ export function useSupporterSubscription(account: string) {
   const [supporter, setSupporter] = useState<SupporterData | null>(null)
   const [loading, setLoading] = useState(true)
 
-  const rawSupporterRef = useRef<any>(null)
+  const rawSupporterRef = useRef<unknown>(null)
   const supporterSubscription = useRef<any>(null)
 
   const onSupporterHandler = useCallback((err, supporter) => {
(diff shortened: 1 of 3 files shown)


src/hooks/useSubscriptions.ts
```diff
@@ -41,18 +41,17 @@ export function useConfigSubscription(garden: Garden | null) {
 
     configSubscription.current = garden.onConfig(onConfigHandler)
 
-    return () => {
-      configSubscription.current = null
-    }
+    return () => configSubscription.current.unsubscribe()
```
reznord, line 44: This should be:

```suggestion
    return () => {
      configSubscription.current.unsubscribe()
    }
```

because useEffect expects undefined during unsubscribing.

src/hooks/useSubscriptions.ts
```diff
@@ -41,18 +41,17 @@ export function useConfigSubscription(garden: Garden | null) {
...
   }, [garden, onConfigHandler])
 
   return config
 }
 
 export function useProposalsSubscription(filters: FiltersType) {
   const { config, connector } = useGardenState()
-  const [proposals, setProposals] = useState<any[]>([])
-  const proposalsSubscription = useRef(null)
+  const [proposals, setProposals] = useState<any>([])
```
reznord, line 52: why did we change the type of the proposals here and generalised it to `any`?

src/hooks/useSubscriptions.ts
```diff
@@ -84,9 +83,7 @@ export function useProposalsSubscription(filters: FiltersType) {
       onProposalsHandler
     )
 
-    return () => {
-      proposalsSubscription.current = null
-    }
+    return () => proposalsSubscription.current.unsubscribe()
```
reznord, line 86: This should be:

```suggestion
    return () => {
      proposalsSubscription.current.unsubscribe()
    }
```
because useEffect expects undefined during unsubscribing.

src/hooks/useSubscriptions.ts
```diff
@@ -144,9 +141,7 @@ export function useProposalSubscription(
       onProposalHandler
     )
 
-    return () => {
-      proposalSubscription.current = null
-    }
+    return () => proposalSubscription.current.unsubscribe()
```
reznord, line 144: This should be:

```suggestion
    return () => {
      proposalSubscription.current.unsubscribe()
    }
```

because useEffect expects undefined during unsubscribing.

src/logic/garden-logic.ts
```diff
@@ -15,12 +15,8 @@ export default function useGardenLogic() {
...
 
   const actions = useActions()
-  const [
-    proposals,
-    filters,
-    proposalsFetchedCount,
-    blockHasLoaded,
-  ] = useProposals()
+  const [proposals, filters, proposalsFetchedCount, blockHasLoaded] =
+    useProposals()
```
reznord, line 19: since there's no change in here, can we revert the changes in this file?

src/hooks/useSubscriptions.ts
```diff
@@ -110,8 +107,8 @@ export function useProposalSubscription(
   const [proposal, setProposal] = useState(null)
   const [loading, setLoading] = useState(true)
 
-  const rawProposalRef = useRef<unknown>(null)
-  const proposalSubscription = useRef<unknown>(null)
+  const rawProposalRef = useRef<any>(null)
+  const proposalSubscription = useRef<any>(null)
```
reznord, line 111: why did we change the type from `unknown` to `any`? TS recommends to use `unknown` when you don't know what are the types of the attributes.

So, can we revert it to `unknown` like it was earlier?

reznord (changes requested):
Left some suggestions and questions as well.

src/hooks/useSubscriptions.ts
tmoutinho: Done ✅

src/hooks/useSubscriptions.ts
tmoutinho: Done ✅

src/logic/garden-logic.ts
tmoutinho: It was auto formatting thing 😅

reznord (approved):
LGTM! 👍🏻

0xGabi (approved):
LGTM!

PR #781, opened 2022-01-07
chore: fix missing source-maps & reduce bundle size

reznord:
#### Context

Earlier in #772 , I have disabled source-maps temporarily for suppressing the warnings in the console.

#### Problem Statement

But that disabled source-maps overall for the entire gardens project and it is a bit hard to debug UI issues since it will point to the transpiled code.

#### Solution
This PR fixes the missing source-maps warnings and enables source-maps which would allow us for easy debugging over-all.

Also, apart from that, I installed `fork-bn.js` and aliased all the `bn.js` chunks in the bundle to the `fork-bn.js`. This allowed us to trim off ~46kB from the main bundle.

#### Note:

Please follow the below mentioned steps for testing:

- Remove `node_modules` & `build` folders: `rm -rf node_modules build`
- Install depenedencies `yarn install` or `yarn`
- Run the project in development / production mode.

##### Before:
image

##### After:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "yarn run sync-assets && GENERATE_SOURCEMAP=false react-app-rewired start",
-    "build": "yarn run sync-assets && GENERATE_SOURCEMAP=false react-app-rewired build",
+    "start": "yarn run sync-assets && react-app-rewired start",
+    "build": "yarn run sync-assets && react-app-rewired build",
     "lint": "eslint ./src",
     "sync-assets": "copy-aragon-ui-assets ./public",
     "bundlewatch": "bundlewatch",
@@ -28,6 +28,7 @@
     "crypto-browserify": "^3.12.0",
     "dayjs": "^1.8.34",
     "ethers": "^5.5.1",
+    "fork-bn.js": "^4.11.8",
     "https-browserify": "^1.0.0",
     "is-ipfs": "^2.0.0",
     "os-browserify": "^0.3.0",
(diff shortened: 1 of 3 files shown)


tmoutinho (approved):
LGTM!

package.json
```diff
@@ -28,6 +28,7 @@
     "crypto-browserify": "^3.12.0",
     "dayjs": "^1.8.34",
     "ethers": "^5.5.1",
+    "fork-bn.js": "^4.11.8",
```
tmoutinho, line 31: What does this do?

reznord: Bumped into this when I was running the bundle analyzer where it showed multiple chunks of `bn.js` library (an internal dependency of some other modules) in the bundle.

When searched for different ways to remove the duplicates of the `bn.js`, I bumped into [this issue](https://github.com/ChainSafe/web3.js/issues/1178) where they were also facing the same problem as me because of duplicate chunks of `bn.js`.

And in [this comment](https://github.com/ChainSafe/web3.js/issues/1178#issuecomment-464617646) on the same issue, there was a suggestion to install this `fork-bn.js` and alias it in the webpack config.

That helped me in reducing the bundlesize by 46kb.

tmoutinho: Cool! Never heard of this, but nice catch.

kamikazebr (approved):

package.json
reznord: Found the sourcecode of the [`fork-bn.js`](https://github.com/BigMurry/bn.js).

The only difference to the main library is that he changed the name of the package to `fork-bn.js` and re-published it. When you see the [diff](https://github.com/indutny/bn.js/compare/master...BigMurry:master) to the main repo, it is nothing but change in the name of the package.

The reason for deduction: since we have added this package as a direct dependency, webpack aliases all the `bn.js` chunks to `fork-bn.js` (magic of webpack aliases 😛)

PR #786, opened 2022-01-10
GardenNameResolver: disable ethers ens resolver

rperez89:
Left the reasons wrote on the code. Left code commented to re enable once ethers 6 is out

Closes <https://github.com/1Hive/gardens/issues/505>

### DIFF
--- a/src/hooks/useGardenNameResolver.ts
+++ b/src/hooks/useGardenNameResolver.ts
@@ -1,30 +1,41 @@
 import { useEffect, useState } from 'react'
+import {useGardens} from '@providers/Gardens'
 import { useWallet } from '@providers/Wallet'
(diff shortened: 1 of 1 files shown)


kamikazebr:
> 

I understand but you will need that later anyway, create branch, merge etc, not? The benefict to do now is clean code while the release not come, or even if that feature change the code will be clear anyways.

Just PoV: I dont like see commented code in the codebase.

BTW For me test, its just try some name already exist on creatin of garden right?

rperez89:
> d that later anyway, cr

don't see any good reason, since plus you need to remember the branch name where you saved it, search for it etc. honestly doesn't worth for me

tmoutinho (approved):

0xGabi (approved):

PR #788, opened 2022-01-13
Fix: Button calling the Delegate Modal

kamikazebr:
Solves that - https://github.com/1Hive/gardens/issues/484

### DIFF
--- a/src/components/Garden/Home.js
+++ b/src/components/Garden/Home.js
@@ -210,6 +210,7 @@ const Home = function Home() {
                           onRequestNewProposal={handleRequestNewProposal}
                           onWrapToken={handleWrapToken}
                           onUnwrapToken={handleUnwrapToken}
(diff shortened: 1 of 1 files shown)


src/components/Garden/Home.js
```diff
@@ -210,6 +210,7 @@ const Home = function Home() {
                           onRequestNewProposal={handleRequestNewProposal}
```
tmoutinho, line 210: We could try to migrate this file to typescript. WDYT?

kamikazebr: Yah I'll love to do that, i just not do that because the file its huge and that PR need just small amout the code to fix. 

I tried now and need convert others files too. So idk if it worth for that PR.

tmoutinho (approved):

0xGabi (approved):
Good catch!

PR #789, opened 2022-01-13
Fix: LocalStorage exceed Onboarding 

kamikazebr:
It solves - https://github.com/1Hive/gardens/issues/498

That PR fix the problem accumulating and never removing onboarding state, but need user clear your cache in case it get full for another reason.

### Update
With the last commit - https://github.com/1Hive/gardens-ui/pull/789/commits/7e71b116a657a34470343ba0b9339423df5c3ed8
Its possible get better handling errors, but it not solve the storage full.

### Suggestion to improvement
The suggestion will be pop-up some button to click and clear localStorage.
The question is: Has some localstorage key we should keep?

### DIFF
--- a/src/components/Onboarding/hooks/useProgressSaver.ts
+++ b/src/components/Onboarding/hooks/useProgressSaver.ts
@@ -8,7 +8,10 @@ import {
   setItem,
 } from '../utils/progress-utils'
 
-export default function useProgressSaver(onConfigChange, onStepChange) {
+export default function useProgressSaver(
+  onConfigChange: any,
+  onStepChange: any
+) {
   const { account, chainId } = useWallet()
   const [resumed, setResumed] = useState(false)
 
@@ -21,7 +24,7 @@ export default function useProgressSaver(onConfigChange, onStepChange) {
   }, [])
 
   const handleSaveConfig = useCallback(
-    config => {
+    (config) => {
       const progress = getItem(account, chainId)
       setItem(account, chainId, { ...progress, config })
     },
@@ -29,7 +32,7 @@ export default function useProgressSaver(onConfigChange, onStepChange) {
   )
 
   const handleSaveStep = useCallback(
-    step => {
+    (step) => {
       const progress = getItem(account, chainId)
       setItem(account, chainId, { ...progress, step })
     },
--- a/src/components/Onboarding/utils/progress-utils.ts
+++ b/src/components/Onboarding/utils/progress-utils.ts
@@ -1,19 +1,19 @@
 import { getNetworkType } from '@utils/web3-utils'
 import { dataURLtoFile, textToFile } from '@utils/kit-utils'
 
-const getStorageKey = (account, chainId) =>
+const getStorageKey = (account: string, chainId: number) =>
   `onboarding:${getNetworkType(chainId)}:${account}`
 
-export const getItem = (account, chainId) => {
+export const getItem = (account: string, chainId: number) => {
   const item = window.localStorage.getItem(getStorageKey(account, chainId))
   return item ? JSON.parse(item) : null
 }
 
-export const removeItem = account => {
-  window.localStorage.removeItem(getStorageKey(account))
+export const removeItem = (account: string, chainId: number) => {
+  window.localStorage.removeItem(getStorageKey(account, chainId))
 }
 
-export const setItem = (account, chainId, item) => {
+export const setItem = (account: string, chainId: number, item: any) => {
   window.localStorage.setItem(
     getStorageKey(account, chainId),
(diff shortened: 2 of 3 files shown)


src/components/Onboarding/utils/progress-utils.ts
```diff
@@ -1,19 +1,19 @@
...
 
-export const getItem = (account, chainId) => {
+export const getItem = (account: string, chainId: number) => {
   const item = window.localStorage.getItem(getStorageKey(account, chainId))
   return item ? JSON.parse(item) : null
 }
 
-export const removeItem = account => {
-  window.localStorage.removeItem(getStorageKey(account))
+export const removeItem = (account: string, chainId: number) => {
```
kamikazebr, line 12: The problem happens here. Its called removeItem passing chainID but the method just accepted one argument,so chainID was undefined.

Migratiing to TS it show on compile time.

src/components/Onboarding/hooks/useProgressSaver.ts
```diff
@@ -8,7 +8,10 @@ import {
   setItem,
 } from '../utils/progress-utils'
 
-export default function useProgressSaver(onConfigChange, onStepChange) {
+export default function useProgressSaver(
+  onConfigChange: any,
+  onStepChange: any
+) {
```
kamikazebr, line 14: It's possible get better types here.

src/components/Onboarding/utils/progress-utils.ts
```diff
@@ -22,7 +22,10 @@ export const setItem = (account, chainId, item) => {
 
 const GARDEN_ASSETS = ['logo_type', 'logo', 'token_logo']
 
-function recoverBlob(file, type = 'dataUrl') {
+function recoverBlob(
+  file?: { content: any; blob: { path: any } },
+  type = 'dataUrl'
+) {
```
kamikazebr, line 28: Its possible extract to named type here or get better type already used someelse.

src/components/Onboarding/utils/progress-utils.ts
```diff
@@ -31,8 +34,11 @@ function recoverBlob(file, type = 'dataUrl') {
   return null
 }
 
-export function recoverAssets(config) {
-  GARDEN_ASSETS.forEach(assetType => {
+export function recoverAssets(config: {
+  garden: { [x: string]: any }
+  agreement: { covenantFile: any }
```
kamikazebr, line 39: Same here, its possible with more investigation on code get better types.

src/utils/kit-utils.js
```diff
@@ -58,7 +58,7 @@ export const readFile = (reader, file) => {
 export function dataURLtoFile(dataurl, filename) {
   const arr = dataurl.split(',')
   const mime = arr[0].match(/:(.*?);/)[1]
-  const bstr = atob(arr[1])
+  const bstr = window.atob(arr[1])
```
kamikazebr, line 61: Its show `atob` like deprecated method, but the nodejs version its deprecated not the DOM one.

So using window solves and avoid possible future bugs.

src/components/Onboarding/utils/progress-utils.ts
```diff
@@ -1,28 +1,35 @@
...
-  )
+export const setItem = (account: string, chainId: number, item: any) => {
+  try {
+    window.localStorage.setItem(
+      getStorageKey(account, chainId),
+      JSON.stringify(item)
+    )
+  } catch (error) {
+    throw new Error('Disk space exceed. Clear your cache.')
+  }
```
kamikazebr, line 24: More info on PR description - https://github.com/1Hive/gardens-ui/pull/789

tmoutinho (approved):

src/components/Onboarding/utils/progress-utils.ts
0xGabi: I wonder if a good place to clear the cache would be when the user clicks that don't want to continue from previous progress:

https://user-images.githubusercontent.com/9082013/150039306-445d1ecd-d707-4933-bd16-2ebf08c50f71.mov

0xGabi (approved):
LGTM!

kamikazebr:
Idr but i think it already remove when click in No. But it too not solve the full storage problem. To solve that we need check which storage keys we cant remove and display some popup in case the error full storage ask to users accept clean it up. It should be on new PR. I should create it?

src/components/Onboarding/utils/progress-utils.ts
kamikazebr: Ohh!! Nice catch! Yah new PR for that could be great and it easy fix will avoid full localstorage mainly in users do having tests.

PR #794, opened 2022-01-13
Upgrade to EVMcrispr v0.3.4

sembrestels:
There are many fixes in the last two releases that affect both EVM Executor and EVM Terminal:

New features:

* New command that introduces the ability to upgrade apps.
* You can pass strings with spaces now to the console using double-quote notation.
* It now supports time units in numeric parameters (such as 1s for 1 second, 1m for 1 minute, 1h for 1 hour...).

Fixes:

* It now recognizes fixed-size array parameters, needed in some agent calls.
* It now loads DAO apps even if some of them don’t have a reachable artifact.
* It now recognizes properly "false" as a false boolean in exec, act, and install commands.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@1hive/connect-agreement": "^0.9.5",
     "@1hive/connect-gardens": "^0.1.12",
     "@1hive/connect-react": "^0.9.9",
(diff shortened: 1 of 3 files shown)


reznord (approved):
LGTM

0xGabi (approved):

PR #797, opened 2022-01-15
Set token logo as wrappableToken if brocolli selected

tmoutinho:
https://github.com/1Hive/gardens-ui/blob/staging/src/services/github.js#L22-L31
https://github.com/1Hive/dao-list/blob/master/xdai.json#L328-L330

Captura de ecrã 2022-01-19, às 19 35 48


src/components/Onboarding/Screens/GardenMetadata.js
```diff
@@ -360,6 +365,23 @@ function GardenMetadata() {
...
+                <div
+                  css={`
+                    width: 100%;
+                  `}
+                >
+                  <FileUploaderField
+                    allowedMIMETypes={['image/jpeg', 'image/png']}
+                    file={formData[WRAPPABLE_LOGO]}
+                    id="file-uploader-3"
+                    label="ERC20 LOGO"
```
0xGabi, line 378:
```suggestion
                    label="EXISTING TOKEN ICON"
```

0xGabi: I wonder to keep consistency between the two token assets names. We can call the one above: `GARDEN TOKEN ICON`.

src/services/github.js
```diff
@@ -28,7 +28,17 @@ export async function publishNewDao(daoAddress, daoMetadata, chainId) {
...
         `${ASSETS_FOLDER_BASE}/${daoMetadata.name}/token_logo.${daoMetadata.token_logo.imageExtension}`,
-    })
+    }
+
+    // if case it's a Boboli type we set the wrappableToken
+    if (daoMetadata.wrappable_token_logo !== null) {
+      newDaoItem.wrappableToken = {
+        logo: `${ASSETS_FOLDER_BASE}/${daoMetadata.name}/wrappable_token_logo.${daoMetadata.wrappable_token_logo.imageExtension}`,
+      }
+    }
```
0xGabi, line 38: @kamikazebr we are going to need to include this in the `dao-list-middleware` right?

0xGabi (changes requested):
LGTM!

src/services/github.js
kamikazebr: Just on github.ts file we need added it after. @0xGabi

src/components/Onboarding/Screens/GardenMetadata.js
tmoutinho: Already changed this 👌

0xGabi (approved):
LGTM!

src/components/Onboarding/Screens/LaunchGarden.tsx
```diff
@@ -1,17 +1,21 @@
...
 
   useEffect(() => {
     const onboardingWizardHeaderEl = document.getElementById(
       'onboarding-wizard-header'
     )
 
-    onboardingWizardHeaderEl.scrollIntoView({
-      behaviour: 'smooth',
+    onboardingWizardHeaderEl?.scrollIntoView({
+      behavior: 'smooth',
```
kamikazebr, line 18: nice catch!

src/services/github.ts
```diff
@@ -39,7 +38,19 @@ export async function publishNewDao(
       token_logo: daoMetadata.token_logo
         ? `${ASSETS_FOLDER_BASE}/${daoMetadata.name}/token_logo.${daoMetadata.token_logo.imageExtension}`
         : null,
-    })
+    }
+
+    newDaoList.push(
+      daoMetadata.wrappableToken === null
```
kamikazebr, line 44: Here and in the line 295 we have 2 differnts outcomes.

Check that tests about undefined and null 

image

On GithubMetadata we will set null to wrappableToken or undefined?

The checks are differents.

I'll suggest is keep the pattern, setting null so we need change the types, removing the optional and putting | null on wrappableToken

image


And add the default null - line 32
image

I could commit that changes if needed.

tmoutinho: Accordingly to Gabi, the wrappableToken should not be sent

kamikazebr (approved):

PR #812, opened 2022-01-26
Feature: Script Check Env

kamikazebr:
This new feature will prevent us from wasting time looking for a bug, when sometimes all that is missing is a variable enviroment

The script it's based on **.env.example** file to check if the env its **required**

### Example of Fail

image

### Example of Success

image

awesome-minions

### Using
It were attached to `yarn start`

### Future Future possible Improvement
- Automatic clone .env.example with default values to .env
- Add in build script and make modifications to throw errors too in Vercel

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Gardens UI  <a href="https://gardens.1hive.org/"><img align="right" src=".github/assets/gardens.svg" height="80px" /></a>
+# Gardens UI <a href="https://gardens.1hive.org/"><img align="right" src=".github/assets/gardens.svg" height="80px" /></a>
 
 ## Gardens are secure digital economies run by the community members that make them valuable.
 
@@ -7,31 +7,24 @@
 ## How to run locally
 
 #### Clone the repository
+
 ```
 git clone https://github.com/1Hive/gardens-ui.git
 ```
 
 #### Install dependencies
+
 ```
 cd gardens-ui
 yarn
 ```
 
 #### Start UI
+
 ```
 yarn start
 ```
 
 ### Environment variables
-```
-REACT_APP_GITHUB_API_TOKEN # API token for GitHub used to fetch Gardens metadata.
-REACT_APP_PINATA_API_TOKEN # API token for Pinata used to pin IPFS content.
-REACT_APP_POLYGON_ETH_NODE # Env to set a custom node on Polygon.
-REACT_APP_XDAI_ETH_NODE # Env to set a custom node on xDai.
-REACT_APP_RINKEBY_ETH_NODE # Env to set a custom node on Rinkeby.
-REACT_APP_PORTIS_ID # App ID to setup Portis provider.
-REACT_APP_POCKET_API_KEY # API key for Pocket to setup a ethers default provider connected to multiple nodes.
-REACT_APP_ALCHEMY_API_KEY # API key for Alchemy to setup a ethers default provider connected to multiple nodes.
-REACT_APP_ETHERSCAN_API_KEY # API key for Etherscan to setup a ethers default provider connected to multiple nodes.
-REACT_APP_INFURA_API_KEY # API key for Infura to setup a ethers default provider connected to multiple nodes.
-```
+
+Check out the [.env.example file](.env.example)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -7,31 +7,24 @@
...
-REACT_APP_XDAI_ETH_NODE # Env to set a custom node on xDai.
-REACT_APP_RINKEBY_ETH_NODE # Env to set a custom node on Rinkeby.
-REACT_APP_PORTIS_ID # App ID to setup Portis provider.
-REACT_APP_POCKET_API_KEY # API key for Pocket to setup a ethers default provider connected to multiple nodes.
-REACT_APP_ALCHEMY_API_KEY # API key for Alchemy to setup a ethers default provider connected to multiple nodes.
-REACT_APP_ETHERSCAN_API_KEY # API key for Etherscan to setup a ethers default provider connected to multiple nodes.
-REACT_APP_INFURA_API_KEY # API key for Infura to setup a ethers default provider connected to multiple nodes.
-```
+
+Check out the [.env.example file](.env.example)
```
kamikazebr, line 30: To keep consistency

package.json
```diff
@@ -3,12 +3,13 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "yarn run sync-assets && react-app-rewired start",
+    "start": "yarn run check-env && yarn run sync-assets && react-app-rewired start",
```
kamikazebr, line 6: Run before yarn start, so we can catch if required ENvs its setup

package.json
```diff
@@ -103,6 +104,7 @@
     "react-scripts": "^5.0.0",
     "rimraf": "^2.6.2",
     "svgo": "^2.8.0",
+    "ts-node": "^10.4.0",
```
kamikazebr, line 107: it's in devDependency.

scripts/read-env-file.ts
```diff
@@ -0,0 +1,160 @@
...
+  return result
+}
+
+function normalizePath(path?: string): string {
+  path = path || resolve(process.cwd(), '.env')
+  return isAbsolute(path) ? path : resolve(process.cwd(), path)
+}
+
+function checkInvalidFormat(key: string, value: string, index: number) {
+  const undef = !key
```
kamikazebr, line 37: Copied from read-env-file npm package and changed here to be able get key without values

tmoutinho (approved):
Super cool 🤌

scripts/read-env-file.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (line.length === 0 || line[0] === '#') return
+
+    const splitLine = line.split('=')
+    const key = (splitLine[0] || '').trim()
+    const value = (splitLine.slice(1).join('=') || '').trim()
+
+    /* Check some error conditions and throw if met */
+    checkInvalidFormat(key, value, index)
+
+    /* If value was quoted, the asigned value should not invlude the quotes */
```
0xGabi, line 67:
```suggestion
    /* If value was quoted, the asigned value should not include the quotes */
```

scripts/read-env-file.ts
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * @description Given an array of paths, will read each and add the key/value
+ * pairs to an object which is returned at the end. Any re-used keys will be overwritten by
+ * the most recent value.
+ * @param {string[]} paths An array of file paths to read.
+ * @returns {Promise<EnvVars>} An aggregate object of key/value pairs
+ * @throws Error if any file is not in valid format or not found
+ */
+export async function readMultiple(paths: string[]): Promise<EnvVars> {
```
0xGabi, line 133: I am not sure if we should include this method if we are not yet using it. What was your rationale to leave it?

0xGabi (approved):
Great work!

scripts/read-env-file.ts
kamikazebr: I was just to remove it now. In fact i think i'll put that in npm package in new repo, because its could be util in another project of 1Hive, WDYT?

I was with doubt about it, because it will avoid need use the whole package to use that function, but i think its better remove.

kamikazebr: Removed 👯

PR #813, opened 2022-01-26
[Feature] Add Discord url pattern to forumURL at Create Proposal

kamikazebr:
The Regexp added to threat **forum.1hive.org** and **discord channels/threads**

Just allowing HTTPS and url ending with slash.

Running Tests:

image

In Create Proposal screen:
image

~Note: The previous implementation used the `forumURL` setted in Onboarding Screen as reference to Create Proposal.~

- Integrated the `forumUrl` to regexp together with discord regexp

### DIFF
--- a/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
+++ b/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
@@ -62,7 +62,10 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
   const fundingMode = formData.proposalType === FUNDING_PROPOSAL
 
   const forumRegex = new RegExp(
-    /https:\/\/(forum\.1hive\.org|discord\.com\/channels)\//
+    String.raw`(${connectedGarden.forumURL.replace(
+      /[-/\\^$*+?.()|[\]{}]/g,
+      '\\$&'
+    )}|https:\/\/discord\.com\/channels\/)`
   )
 
   const handleAmountEditMode = useCallback(
--- a/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
+++ b/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
@@ -61,11 +61,14 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
 
   const fundingMode = formData.proposalType === FUNDING_PROPOSAL
 
+  // Escaping forumURL to avoid misuse with regexp
+  const forumURLRegex = connectedGarden.forumURL.replace(
+    /[-/\\^$*+?.()|[\]{}]/g,
+    '\\$&'
+  )
+
   const forumRegex = new RegExp(
-    String.raw`(${connectedGarden.forumURL.replace(
-      /[-/\\^$*+?.()|[\]{}]/g,
-      '\\$&'
-    )}|https:\/\/discord\.com\/channels\/)`
+    String.raw`(${forumURLRegex}|https:\/\/discord\.com\/channels\/)`
   )
 
   const handleAmountEditMode = useCallback(
--- a/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
+++ b/src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
@@ -20,6 +20,7 @@ import { useMultiModal } from '@components/MultiModal/MultiModalProvider'
 import { usePriceOracle } from '@hooks/usePriceOracle'
 import BigNumber from '@lib/bigNumber'
 import { toDecimals } from '@utils/math-utils'
+import { escapeRegex, regexToCheckValidProposalURL } from '@utils/regex-utils'
 import { formatTokenAmount, isStableToken } from '@utils/token-utils'
 import { calculateThreshold, getMaxConviction } from '@lib/conviction'
 
@@ -62,13 +63,8 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
   const fundingMode = formData.proposalType === FUNDING_PROPOSAL
 
   // Escaping forumURL to avoid misuse with regexp
-  const forumURLRegex = connectedGarden.forumURL.replace(
-    /[-/\\^$*+?.()|[\]{}]/g,
-    '\\$&'
-  )
-
-  const forumRegex = new RegExp(
-    String.raw`(${forumURLRegex}|https:\/\/discord\.com\/channels\/)`
+  const forumRegex = regexToCheckValidProposalURL(
+    escapeRegex(connectedGarden.forumURL)
   )
 
   const handleAmountEditMode = useCallback(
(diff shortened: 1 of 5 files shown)


src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -61,11 +61,13 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
 
   const fundingMode = formData.proposalType === FUNDING_PROPOSAL
 
-  const forumRegex = new RegExp(connectedGarden.forumURL)
```
0xGabi, line 64: Note that this was the `forumURL` because Gardens can specify their own forum. So we need to include a regEx that uses the `forumURL` as variable.

0xGabi (changes requested):

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
kamikazebr: Right. Idk how it was working without escape the string, it will just blow up because the double slash on https:// it's nots escaped in orignal forumURL.

Anyway, I'll add them back mergeing with the discord

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -61,11 +61,16 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
...
+  const forumRegex = new RegExp(
+    String.raw`(${connectedGarden.forumURL.replace(
+      /[-/\\^$*+?.()|[\]{}]/g,
+      '\\$&'
+    )}|https:\/\/discord\.com\/channels\/)`
+  )
```
kamikazebr, line 69: Escaping the forumURL to avoid getting regex from Onboarding.

Should I put it all in consts btw?
Or its possible to understand?

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -61,11 +61,19 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
 
   const fundingMode = formData.proposalType === FUNDING_PROPOSAL
 
-  const forumRegex = new RegExp(connectedGarden.forumURL)
+  // Escaping forumURL to avoid misuse with regexp
+  const forumURLRegex = connectedGarden.forumURL.replace(
```
kamikazebr, line 65: I'm assuming forumURL it's required, will always be there, its correct?

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -178,7 +189,7 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
     }
 
     if (link && !forumRegex.test(link)) {
-      errors.push('Forum post link not provided ')
+      errors.push('Forum post link not provided or not valid')
```
kamikazebr, line 192: Make sense that addition?

tmoutinho:
It would be nice to have a test for this.

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
tmoutinho: What if the connectedGarden.forumURL is null? But if it is required then there is no need for more validation 👌

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -61,11 +61,19 @@ const AddProposalPanel = React.memo(({ setProposalData }) => {
 
   const fundingMode = formData.proposalType === FUNDING_PROPOSAL
 
-  const forumRegex = new RegExp(connectedGarden.forumURL)
+  // Escaping forumURL to avoid misuse with regexp
+  const forumURLRegex = connectedGarden.forumURL.replace(
+    /[-/\\^$*+?.()|[\]{}]/g,
+    '\\$&'
```
tmoutinho, line 67: Also can we create a const for this regex? Maybe it will be necessary to use it elsewhere.

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
willjgriff: We should stop using the term "Forum post" and start using "Proposal information". 

This should then read "Proposal information link not provided".

willjgriff:
We use the term "Forum Post" in a bunch of places. Ideally we would change these to "Proposal Information" as it's more generic. Can you see how easy it would be to update it everywhere it needs to be to "Proposal Information"? If it's too involved for some reason leave it for now and change it if this feature is more widely used. 

@0xGabi wdyt?

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
kamikazebr: Nice idea. Should we build some simply function and put where? I'll check and do that.

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
kamikazebr: It mislead the user, Say its not provided while the text input have alot text, have sure @willjgriff ?

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
kamikazebr: I extracted to allow make the test file easily. have look @tmoutinho

kamikazebr:
> We use the term "Forum Post" in a bunch of places. Ideally we would change these to "Proposal Information" as it's more generic and can encapsulate the Discord link too. Can you see how easy it would be to update it everywhere it needs to be to "Proposal Information"? If it's too involved for some reason leave it for now and we can change it if this feature is more widely used.

Looking in the code superficially, seems be have some places to change, but should be effort 4 task to do, but i'll recommend open in new PR because it will more easy to track the changes with screenshots.

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
kamikazebr: Fixed

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
tmoutinho: Nice 👍🏼

tmoutinho (approved):
✌️

0xGabi (approved):
LGTM!

0xGabi:
@willjgriff I agree with @kamikazebr let's do the changes on a new PR 👍

PR #814, opened 2022-01-27
Fix: small UI fixes

0xGabi:
Before:
Screen Shot 2022-01-26 at 9 33 11 PM

After:
Screen Shot 2022-01-26 at 9 32 57 PM


tmoutinho (approved):
✌️

PR #825, opened 2022-02-10
Update wallet behavior 

tmoutinho:
- [x] If in supported network it should be connected in the first place
- [x] `WrongNetwork` Button should be active on refresh
- [x] When you change network on metamask it doesn't prompt you to switch network ( Instead update the front-end with a button to switch the network if you would like )
- [x] Make the dropdown on the header like gnosis-safe
- [x] Refactor Wallet provider since sometimes it works sometimes it doesn't.

https://user-images.githubusercontent.com/13184907/154695544-d24c6930-bd76-4b89-b2df-cacd6691a2fd.mov

Captura de ecrã 2022-02-16, às 21 00 50

Captura de ecrã 2022-02-16, às 21 01 08

Captura de ecrã 2022-02-18, às 13 52 15

Captura de ecrã 2022-02-24, às 22 52 40

Captura de ecrã 2022-02-24, às 23 07 31

Captura de ecrã 2022-03-01, às 16 12 36


src/components/Account/AccountModule.js
```diff
@@ -79,10 +83,10 @@ function AccountModule({ compact }) {
 
   const { direction, screenIndex } = useMemo(() => {
     const screenId = (() => {
+      if (account) return 'connected'
```
kamikazebr, line 84: Changing the other not will avoid show error screen when it happen?

src/components/Account/AccountModule.js
```diff
@@ -122,6 +126,21 @@ function AccountModule({ compact }) {
     }
   }, [screenId])
 
+  useEffect(() => {
+    if (window?.ethereum !== undefined) {
+      window?.ethereum.on('chainChanged', (newChainId) => {
+        console.log(`chainChanged`, newChainId)
```
kamikazebr: Some console.log

src/components/Account/AccountModule.js
```diff
@@ -40,21 +41,21 @@ function AccountModule({ compact }) {
...
-        console.log('error ', error)
-      }
-    },
-    [connect]
-  )
+  const activate = async (providerId) => {
+    try {
+      await connect(providerId)
+    } catch (error) {
+      console.log('error ', error)
```
kamikazebr, line 56: That need be logged? if yes, maybe console.error()?

src/components/Account/Buttons/ChangeNetworkModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+  compact,
+  visible,
+  onClose,
+}: ChangeNetworkModalProps) {
+  const { width } = useViewport()
+
+  const handleSwitch = async () => {
+    try {
+      await switchNetwork(100)
+      onClose()
```
kamikazebr, line 28: if you `throw new Error('some error test')` it show some popscreen error? the default one with sad flower? Just curious

src/networks.js
```diff
@@ -170,6 +172,12 @@ export const switchNetwork = async (chainId) => {
   }
 }
 
+export function isSupportedConnectedChain() {
+  if (window?.ethereum === undefined) return true
```
kamikazebr, line 176: It not should return `undefined` instead `true`?

Looks like the `isSupportedConnectedChain` check and should return `true` if the current chain is connected, right?

Addind `boolean | undefined` open the possibility of threat that behaviour, where the window.ethereum its not loaded correctly, maybe in mobile or browser without metamask, make sense?

I did not test nothing there, just looking in the code and trying understand the behaviou.

What you think?

src/networks.js
```diff
@@ -188,8 +196,6 @@ export function getAgreementConnectorConfig(chainId) {
   }
 }
 
-export const SUPPORTED_CHAINS = [4, 100, 137] // Add  arbitrum  chains id + fill the network json with the data
-
 export function isSupportedChain(chainId) {
```
kamikazebr, line 199: It's possible translate that file to TS?

kamikazebr:
Awesome job!!! <3

src/components/Account/AccountModule.js
tmoutinho: This was already there I didn't change 🤷🏼‍♂️

src/components/Account/AccountModule.js
tmoutinho: I will remove this 👌

src/components/Account/AccountModule.js
tmoutinho: I am still checking this. But yes it should not be a console log there 👌

src/components/Account/Buttons/ChangeNetworkModal.tsx
tmoutinho: I don't really know. But I will check it.

src/networks.js
tmoutinho: I need to test this further. But what I was going for it to be careful with mobile that doesn't have access to the window. In that case we don't show the wrong network button, but we keep the enable account one.

tmoutinho: If the function return false then the wrong networks button appears.

src/networks.js
tmoutinho: I can try 👌

src/components/Account/AccountModule.js
kamikazebr: It show to me the order changed, the `if (error)` was in first place.

image

src/networks.js
kamikazebr: I see, and if you return `undefined` instead and threat it like some error and use the default buttons?

src/components/Account/Buttons/ChangeNetworkModal.tsx
tmoutinho: Basically the throw new Error() appears like this:

Captura de ecrã 2022-02-26, às 17 30 48

kamikazebr: So nothing in the screen? We need handle that error, what you think?

src/networks.js
tmoutinho: I tried but it has too many ramifications.

src/components/Account/Buttons/ChangeNetworkModal.tsx
tmoutinho: Just like this. 

Captura de ecrã 2022-02-26, às 18 40 29

src/networks.js
tmoutinho: No. Its has 2 states, if true then its not a supported network, else it's supported. It's simple.

src/components/Account/Buttons/ChangeNetworkModal.tsx
kamikazebr: Awesome! My bad not follow the code inside the method.

src/networks.js
tmoutinho: No Felipe. The function just say if you are in the wrong network ( the chainid is not one of the supported networks ). Thats all. Maybe the name is not the best.

kamikazebr (approved):

kamikazebr:
image
That appear same in the correct network.

ANd i noticed the word "Mainnet" and "gnosis chain" its hardcoded.

PR #834, opened 2022-02-27
Change visualization of stakes in profile

tmoutinho:
Captura de ecrã 2022-03-21, às 23 16 09


src/components/Garden/Feed/Icons.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
```
rperez89, line 1: why are we doing this instead of using the 1hive ui component?

src/components/Garden/Feed/ProposalCreator.tsx
```diff
@@ -1,24 +1,25 @@
...
-} from '@1hive/1hive-ui'
+import { GU, shortenAddress, textStyle, useTheme } from '@1hive/1hive-ui'
 import ProposalIcon from '@components/ProposalIcon'
 
 import { convertToString } from '@/types'
 import { getProfileForAccount } from '@lib/profile'
 import { dateFormat } from '@utils/date-utils'
 import { addressesEqual } from '@utils/web3-utils'
 import { ZERO_ADDR } from '@/constants'
+import { ProposalType } from '@/hooks/constants'
```
rperez89, line 10: i would recommend to move the `constants` file out from the hooks since are types that are being used across the app and rename it to something like `types`

src/components/Garden/Feed/RightPanel.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react'
```
rperez89, line 1: why we removed the useMemo implementation since has a dependency on all this `account, largeMode, mobileMode, tabletMode`, which specially the account one can create a rerender of the whole component?

rperez89:
> Captura de ecrã 2022-02-27, às 21 01 20
> 
> Captura de ecrã 2022-03-06, às 16 30 07

Hey @tmoutinho trying to understand a bit the changes done regarding to visualization but couldn't understand them, in the first image i see that there is a big address that is overflowing the component but that does not happen in production right now.

I think that what you wanted to do is to split the staking by garden right?, first i would change that address for the garden name.

Giving it a try here is what is shown in production from 2 difrerent garden

Screen Shot 2022-03-12 at 11 38 43 PM

and here on your branch 

Screen Shot 2022-03-12 at 11 46 01 PM

so we are missing stakes for others gardens

rperez89 (changes requested):

src/components/Garden/Feed/Icons.js
tmoutinho: Hey @rperez89, since the 1hive-ui types has some problems, migrating some files to typescript is impossible without adding some redundant and unnecessary props to some of the components ( in this case the majority of the icons ). When the PR from Felipe that make this icons optional, this file will be removed. I will add a TODO in this file.

src/components/Garden/Feed/ProposalCreator.tsx
tmoutinho: Sure, I can do that.

src/components/Garden/Feed/RightPanel.tsx
tmoutinho: I don't remember why, but I will add it again.

src/components/Garden/Feed/Icons.js
kamikazebr: I think that Fix make all options its merged some time ago, just need bump the 1hive-ui version.

More details here - https://github.com/1Hive/1hive-ui/pull/28

Version 1.0.17, you can check if that works to you @tmoutinho ?

src/components/Garden/Filters/TextFilter.tsx
```diff
@@ -1,7 +1,13 @@
...
 import { GU, SearchInput, textStyle } from '@1hive/1hive-ui'
 
-const TextFilter = React.memo(({ text, onChange, placeholder = '' }) => {
+type TextFilterProps = {
+  text: string
+  placeholder: string
+  onChange: () => void
+}
+
+const TextFilter = ({ text, onChange, placeholder = '' }: TextFilterProps) => {
```
kamikazebr, line 10: You really need placeholder have default to empty string? It's also mean placeholder its optional? Maybe the type definition need be optional too?

Just curisous, not a request of change.

tmoutinho: Yes sure, it could be optional. Will update it.

src/components/Garden/Feed/Icons.js
tmoutinho: Yes I will check it.

tmoutinho: Upgraded 1hive-ui to v1.0.17, and removed the Icons.js folder. All good.

src/components/Garden/Filters/TextFilter.tsx
tmoutinho: Updated!

src/components/Garden/Feed/RightPanel.tsx
tmoutinho: Updated!

src/components/Garden/Feed/ProposalCreator.tsx
tmoutinho: Updated!

src/components/Garden/Feed/Icons.js
tmoutinho: The file was removed and the 1hive-ui version was upgraded. With the help of @kamikazebr

kamikazebr (approved):

0xGabi:
> so we are missing stakes for others gardens

I agree with @rperez89. The `Supported Proposals` component should be organized per Garden. Something like:
Screen Shot 2022-03-21 at 3 45 13 PM

I also got the following error testing the deployment with a Rinkeby account: 
Screen Shot 2022-03-21 at 3 06 51 PM

src/components/Profile/StakingTokens.tsx
```diff
@@ -0,0 +1,229 @@
...
+}> => {
+  if (stakes === null) return []
+
+  const gardens: Array<any> = []
+  const stakesPerGarden: Array<any> = []
+
+  // get all unique gardens
+  stakes.map((stake: TransformedStakeType) => {
+    const gardenId = stake?.item?.gardenId
+    if (!gardens.includes(gardenId) && gardenId !== null) {
```
reznord, line 101: This could be better right? (just a bit cleaner)

```suggestion
    const gardenId = stake?.item?.gardenId ?? ""
    if (gardenId && !gardens.includes(gardenId)) {
```

src/components/Profile/StakingTokens.tsx
```diff
@@ -0,0 +1,229 @@
...
+  return stakesPerGarden
+}
+
+function StakingTokens({ myStakes }: StakingTokensProps) {
+  const { gardensMetadata } = useGardens()
+  const theme = useTheme()
+  const { below } = useViewport()
+  const compact = below('large')
+
+  const myActiveTokens = useMemo(() => {
```
reznord, line 123: We don't really need `useMemo` here because if the data of the `myStakes` changes, then the `useMemo` is of no use.

And moreover, it is not such heavy computation, so we are just over-complicating things here.

src/components/Profile/StakingTokens.tsx
```diff
@@ -0,0 +1,229 @@
...
+        text-overflow: ellipsis;
+        overflow: hidden;
+        white-space: nowrap;
+
+        ${proposal?.proposalId &&
+        `cursor: pointer; &:hover {
+            background: ${theme.badge.alpha(0.7)}
+          }
+        `}
+      `}
```
reznord, line 218: can we convert this to styled component? #prettyPlease

src/components/Profile/InactiveProposalsStake.tsx
```diff
@@ -24,44 +31,52 @@ function InactiveProposalsStake({ myInactiveStakes }) {
...
-            selectProposal={handleSelectProposal}
-          />
+          <div
+            key={index}
+            css={`
+              display: flex;
+              align-items: center;
+              margin-bottom: ${1 * GU}px;
+            `}
+          >
```
reznord, line 43: can we convert this to styled component? #prettyPlease

reznord (commented):
Left some comments and suggestions.

src/components/Garden/Feed/ProposalCountdown.tsx
```diff
@@ -1,23 +1,28 @@
...
+}
+
+function ProposalCountdown({ proposal }: CountDownProps) {
   if (proposal.hasEnded) {
     return null
   }
 
   if (proposal.status === PROPOSAL_STATUS_DISPUTED_STRING) {
-    return <Tag>Paused</Tag>
+    return <Tag>{"Paused"}</Tag>
```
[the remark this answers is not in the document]
tmoutinho: Unfortunately we cannot. With the new version of `h1ve-ui` it fails deploy.

src/components/Profile/StakingTokens.tsx
tmoutinho: It was already there. I will leave it be.

src/components/Profile/StakingTokens.tsx
```diff
@@ -0,0 +1,229 @@
...
+    if (!myStakes) {
+      return new BigNumber('0')
+    }
+
+    return myStakes.reduce((accumulator, stake) => {
+      return accumulator.plus(stake.amount)
+    }, new BigNumber('0'))
+  }, [myStakes])
+
+  const stakes = useMemo(() => {
```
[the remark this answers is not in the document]
tmoutinho: Same as above. We can do it in a new PR.

src/components/Profile/InactiveProposalsStake.tsx
tmoutinho: It was already there. We can do it in a new PR

src/components/Profile/StakingTokens.tsx
tmoutinho: I will change the condition ✌️

reznord (approved):

src/components/Profile/InactiveProposalsStake.tsx
```diff
@@ -24,44 +31,52 @@ function InactiveProposalsStake({ myInactiveStakes }) {
...
+            <ProposalItem
+              amount={stake.amount}
+              gardenId={stake.proposal.organization.id}
+              proposalId={stake.proposal.id}
+              proposalName={stake.proposal.metadata}
+              compact={compact}
+              selectProposal={handleSelectProposal}
+            />
```
0xGabi, line 51: What do you think of also grouping the inactive proposals per Garden?

src/components/Profile/StakingTokens.tsx
```diff
@@ -0,0 +1,229 @@
...
+}
+
+function displayedStakes(stakes: Array<StakeItem>, total: BigNumber) {
+  return stakesPercentages(
+    stakes.map(({ amount }) => amount),
+    {
+      total,
+      maxIncluded: DISTRIBUTION_ITEMS_MAX,
+    }
+  ).map((stake: any) => ({
```
0xGabi, line 43: Looking at the deployment I think we are calculating the percentages wrong as we are considering all the stakes across every Garden. Instead, I think we should consider the stake percentages per Garden.

src/components/Profile/StakingTokens.tsx
```diff
@@ -0,0 +1,229 @@
...
+  if (myActiveTokens.eq(0)) {
+    return null
+  }
+
+  const colors = [theme.green, theme.red, theme.purple, theme.yellow]
+  const stakesByGarden = getStakesByGarden(gardensMetadata, stakes)
+
+  return (
+    <Box heading="Supported proposals" padding={3 * GU}>
+      <p>Active token distribution</p>
```
0xGabi, line 149:
```suggestion
      <p>Active token distribution per Garden</p>
```

src/components/Profile/StakingTokens.tsx
```diff
@@ -0,0 +1,229 @@
...
+            margin-top: 25px;
+          `}
+        >
+          <Distribution
+            colors={colors}
+            heading={
+              <p
+                css={`
+                  margin-bottom: 10px;
+                  font-size: 16px;
```
0xGabi, line 163: What do you think of giving the Garden name a bit more style? For example:
```
<Tag mode="identifier">
	{gardenName ?? shortenAddress(garden)}
</Tag>

```

0xGabi (changes requested):
We are super close, well done! I left a few comments.

src/components/Profile/StakingTokens.tsx
rperez89: > Looking at the deployment I think we are calculating the percentages wrong as we are considering all the stakes across every Garden. Instead, I think we should consider the stake percentages per Garden.

+1 on this comment just noticed the same

src/components/Profile/InactiveProposalsStake.tsx
tmoutinho: The data in the inactive stake is slightly different, but I can do it. 👌

src/components/Profile/StakingTokens.tsx
tmoutinho: I think we could aim for something diferente since if we have a lot of active stakes and a lot of gardens it would be a lot of scrolling to do. We could have a toggle or modal for this. But as I suggested this could be done in a next PR. For this moment I will add the Tag element ✌️

src/components/Profile/StakingTokens.tsx
tmoutinho: Good catch. I will fix it.

tmoutinho: Captura de ecrã 2022-03-26, às 16 12 42

made the fix

src/components/Profile/StakingTokens.tsx
tmoutinho: WDYT?
Captura de ecrã 2022-03-26, às 16 12 42

src/components/Profile/InactiveProposalsStake.tsx
tmoutinho: One question @0xGabi . should the inactive one be included on the calculation of the percentage?

0xGabi: I think they should as they are part of the token balance the use have available for voting.

src/components/Profile/StakingTokens.tsx
0xGabi: Looking great! I think we can display the Inactive proposals stake the same way: per Garden and using % instead of token amount. What do you think?

src/components/Profile/StakingTokens.tsx
0xGabi: Looks great!

src/components/Profile/StakingTokens.tsx
tmoutinho: Will do it.

src/components/Profile/InactiveProposalsStake.tsx
tmoutinho: I will add the inactive. But they need to have something to distinguish from the active one. What do you suggest? I could add a separator.

src/components/Profile/StakingTokens.tsx
tmoutinho: @0xGabi I have added the inactive and change the background to a gray one. Also change the title of the block to "Token distribution per Garden" since it included "Active" and "Inactive"

Captura de ecrã 2022-03-28, às 11 58 32

0xGabi: Well done, I like how to merge both view!

0xGabi (approved):
Awesome work!

rperez89 (approved):
LGTM!!

PR #840, opened 2022-03-14
feat: migrate to next.js

reznord:
This PR is a migration path from `create-react-app` to `next.js`.

--------------------------

### Major Changes

Moving forward, we have to use the `public` folder for all static assets like `.svg`, `.png`, `.jpg` etc. These will not be included in the main bundle unlike the current CRA application.

This allowed us in reducing the bundle size by 500 kb (earlier it was around 2Mb) and now it is 1.48Mb. 

I will still try to reduce the framework footprint from 1.48Mb to less than 1Mb, but not sure if can be done or not. 

--------------------------

### Changes

- [x] Fix typescript compilation issues related to next.js.
- [x] Fix styling related issues related to styled-components and next.js.
- [x] Migrate routes from using `react-router` to [`pages`](https://nextjs.org/docs/basic-features/pages) folder.
- [x] Move static assets (`.svg`, `.png`, `.jpg` etc.) to public folder.
- [x] Fix the rendering issues in all the pages during server-sider rendering.
- [x] Migrate the default css in `DiscourseComments.css` to styled-components.
- [x] Migrate fonts to the static folder and load them in the [`_document`](https://nextjs.org/docs/advanced-features/custom-document). This will allow us to remove the `copy-argon-ui-assets` command from the scripts.
- [ ] Setup a custom server.js for pre-checking the env variables.
- [ ] Migrate `@/environment.js` to a config folder so, we can access the config through `process.env.....`
- [ ] Cleanup the dependencies in the project

--------------------------

### Need help

- See if we can migrate the storage from local/session storage to cookies in the `use-wallet`?
- Debug layout issues which are blocking the rendering on the server

--------------------------

### Enhancements

- Switch from custom eslint config to [next.js's default eslint](https://nextjs.org/docs/basic-features/eslint) which will run during the build time as well.
- Use [eslint-plugin-diff](https://github.com/paleite/eslint-plugin-diff) for applying the new eslint config only to the files which are modified.

--------------------------

### Deployments

I had to create a new project on vercel for testing these changes related to next.js since the `framework` in the settings is different for the gardens project (currently it is set to `CRA`, but we need the framework to be `next.js`)

You can find the latest deployments on my fork: https://github.com/reznord/gardens-ui/deployments

⚠️ So, when we merge these changes to staging, we need to change the framework setting from `CRA` to `next.js` and then re-run the build by clearing the cache.


package.json
```diff
@@ -3,10 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "yarn run check-env && yarn run sync-assets && react-app-rewired start",
-    "build": "yarn run sync-assets && react-app-rewired build",
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
     "lint": "eslint ./src",
-    "sync-assets": "copy-aragon-ui-assets ./public",
```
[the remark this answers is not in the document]
reznord: Not anymore :) 

Earlier we were using this for loading the fonts which are moved to the project itself right now.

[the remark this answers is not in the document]
tmoutinho: Never mind, you explained in the call :)

package.json
```diff
@@ -3,10 +3,11 @@
   "version": "0.1.0",
```
0xGabi, line 3: 🤩

src/components/Account/ScreenConnected.js
```diff
@@ -33,15 +30,15 @@ function AccountScreenConnected({ providerId, onClosePopover }) {
...
-    history.push(`/profile`)
+    router.push(`/profile`)
     onClosePopover()
-  }, [history, onClosePopover])
+  }, [router, onClosePopover])
 
   const goToStakeManagement = useCallback(() => {
-    const path = buildGardenPath(history.location, 'collateral')
-    history.push(path)
+    const path = buildGardenPath(router, 'collateral')
```
0xGabi, line 38: When we build the path using `buildGardenPath` we don't replace the correct parameters. The final URL looks something like: `[networkType]/garden/[gardenAddress]/proposal/1`.

Instead, the `GardenList` component as we are doing it manually is working fine: https://github.com/reznord/gardens-ui/blob/staging/src/components/GardensList.js#L80-L84

src/hooks/useRouting.ts
```diff
@@ -1,15 +1,8 @@
...
-type ParamsType = {
-  params: {
-    gardenAddress: string
-    networkType: string
-  }
-} | null
+import { useRouter } from 'next/router'
 
 export function useGardenRoute() {
-  const match: ParamsType = useRouteMatch('/:networkType/garden/:gardenAddress')
```
0xGabi, line 11: Can we double check we are still handling the routes correctly?

src/pages/[networkType]/garden/[gardenAddress]/collateral.tsx
```diff
@@ -0,0 +1,3 @@
+import StakeManagement from '@/components/Garden/Stake/StakeManagement'
```
Comment (0xGabi) [inline src/pages/[networkType]/garden/[gardenAddress]/collateral.tsx:1]: From my understanding, I believe we are not getting the URL params right for all these next.js files.

src/components/DiscourseComments.tsx
```diff
@@ -2,12 +2,25 @@ import React, { useState, useEffect } from 'react'
 import { useConnectedGarden } from '@providers/ConnectedGarden'
 import CustomSanitizedHTML from '@/components/SanitizedHTML'
 import env from '@/environment'
```
0xGabi, line 4: Would like to give a final review once I can access a proposal page with comments.

0xGabi (changes requested):
Anup, really awesome changes!

We have a few small details we need to figure out with the routes. I believe it has to do with the URL params, the `useRoute` hook, and the `buildGardenPath` method.

Some of the issues I noted:
- When I try to access a [proposal](https://gardens-ui-nextjs-ke8z3h2qj-1hive.vercel.app/xdai/garden/0x8ccbeab14b5ac4a431fffc39f4bec4089020a155/proposal/86). I got an error.

- Gardens Covenant link not working.

- The deposit manager link inside a Garden home is not working.

- The settings page link inside Garden home is not working.

src/hooks/useRouting.ts
kamikazebr: Maybe create some jest file test?

reznord: Yes, `networkType` and `gardenAddress` params are available for us from the `useRouter` hook from `next/router`.

This will remove the dependency of performing a route match like what we are doing.

src/components/Account/ScreenConnected.js
reznord: Yes, I am removing the whole `buildGardenPath` dependency entirely.

I have migrated almost all the places where we are using the `buildGardenPath` but there are a few places which I am stuck at (will ping you on discord for this).

src/pages/[networkType]/garden/[gardenAddress]/collateral.tsx
Comment (reznord) [reply src/pages/[networkType]/garden/[gardenAddress]/collateral.tsx]: I didn't really understand what you meant here.

Are you asking how next.js will come to know about the `networkType` and `gardenAddress` from the URL? If so, that will be automatically resolved since the `[networkType]` and `[gardenAddress]` are dynamic parameters.

PR #860, opened 2022-05-13
Settings: Garden permissions

rperez89:
Still didn't added the filters since is going to take me some time given the way the connect returns the data for the roles but i can add later in another pr since i want to move to work on the conviction voting v2.

There is one case where the permission has no name for some reason, needs further investigation.

Screen Shot 2022-05-12 at 9 44 26 PM


src/components/Garden/Preferences/Permissions.tsx
```diff
@@ -0,0 +1,118 @@
+/* eslint-disable react/jsx-key */
```
kamikazebr, line 1: Maybe the problem with names don't show up could be related to it. Same if it's not, I'll recommend remove it and set proper keys attributes to react render properly

src/components/Garden/Preferences/GlobalPreferences.js
```diff
@@ -11,23 +11,48 @@ import {
...
 import { useConnectedGarden } from '@providers/ConnectedGarden'
 import { useWallet } from '@/providers/Wallet'
 import { useEsc } from '../../../hooks/useKeyboardArrows'
 
 import AppsAddresses from './AppsAddresses'
 import EVMExecutor from './EVMExecutor'
+import Permissions from './Permissions'
+
+
+const LayoutWrapper = styled(Layout)`
```
tmoutinho, line 25: Nice 🔥

src/components/Garden/Preferences/Permissions.tsx
tmoutinho: I don't think we really need this line also.

tmoutinho (approved):
Good job :fire. just a small comment.

src/components/Garden/Preferences/GlobalPreferences.js
kamikazebr: Yeah loved that too

src/components/Garden/Preferences/GlobalPreferences.js
```diff
@@ -65,17 +90,18 @@ function GlobalPreferences({ compact, onClose, onNavigation, sectionIndex }) {
     getEvmCrispr()
   }, [account, connectedGarden, ethers, isSafari])
 
+  console.log('sectionIndex ', sectionIndex)
```
kamikazebr, line 93: Maybe remove the console log here

src/hooks/useRoles.ts
```diff
@@ -0,0 +1,90 @@
...
+  granteeName: string | undefined,
+  params: Array<any>,
+  roleHash: string
+}
+
+interface RoleProps {
+  appAddress: string,
+  appName: string,
+  appIcon: string,
+  description: string | undefined, // TODO- check why this is happening proabably old ABI
```
kamikazebr, line 26: You are still getting undefined here some times?

src/hooks/useRoles.ts
```diff
@@ -0,0 +1,90 @@
...
+  }, [gardenState])
+
+  const appsWithRolesResolved = usePromise(appsWithPermissions, [], [])
+
+  function isAnyAddress(address: string) {
+    return address === ANY_ADDRESS
+  }
+
+  const rolesWithEntitiesResolved: Array<RoleProps> = appsWithRolesResolved ? appsWithRolesResolved.map((app: any) => {
+    return app.roles.map((role: any) => {
```
kamikazebr, line 57: It's possible we get the correct type here in instead any, it could help track changes in the code below

reznord (approved):

Issue #418, opened 2021-05-12
Dark mode


fabriziovigevani:
https://www.figma.com/file/w97oARETFFr4WyApvdqV5E/%F0%9F%8C%91-Honey-Pot-Dark-Mode-V2?node-id=0%3A1

kamikazebr:
https://www.joshwcomeau.com/react/dark-mode/
Amazing article about how implement it. It works to React too.

0xGabi:
close with #649

PR #870, opened 2022-08-20
Gardens dark mode support with latests staging changes

0xGabi:


0xGabi:
I push the latest changes on #418

PR #877, opened 2022-09-01
feat: handle stream proposals creation

0xGabi:


src/components/Garden/Feed/ProposalInfo.tsx
```diff
@@ -23,6 +26,9 @@ function ProposalInfo({
 }: ProposalInfoProps) {
   const theme = useTheme()
   const { config } = useGardenState()
+  const { flow, loading: loadingFlow } = useSuperfluidCFAv1(
```
rperez89, line 29: wondering what that this `useSuperfluidCFAv1` means and  if we can get a better naming

src/components/Garden/Feed/ProposalInfo.tsx
```diff
@@ -31,12 +37,10 @@ function ProposalInfo({
     requestToken.decimals
   )
 
-  const hideRequestedInfo = Number(proposal.requestedAmount) === 0
-
   return (
     <div onClick={onSelectProposal}>
       <ProposalDescription proposal={proposal} />
-      {proposal.type !== ProposalTypes.Decision && !hideRequestedInfo && (
+      {proposal.type === ProposalTypes.Proposal && (
```
rperez89, line 43: why we changed this here ? before it meant to be for proposal types of proposals and suggestions

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -243,7 +249,11 @@ const AddProposalPanel = ({ setProposalData }) => {
           placeholder="Select proposal type"
           selected={formData.proposalType}
           onChange={handleProposalTypeChange}
-          items={PROPOSAL_TYPES}
+          items={
```
rperez89, line 252: maybe we can get those items before render time ?

src/components/Garden/ModalFlows/CreateProposalScreens/AddProposal.js
```diff
@@ -253,6 +263,8 @@ const AddProposalPanel = ({ setProposalData }) => {
           {formData.proposalType === SIGNALING_PROPOSAL
             ? `Suggestion proposals are used to gather community sentiment for
         ideas or future funding proposals.`
+            : formData.proposalType === STREAM_PROPOSAL
+            ? `Stream proposals create continuous funding to the beneficiary in relation to the support (conviction) accrue.`
```
rperez89, line 267: same here we can compute this data outside the html

src/components/Garden/ProposalDetail/ProposalDetail.js
```diff
@@ -64,6 +65,7 @@ import { ZERO_ADDR } from '@/constants'
 // assets
 import warningIcon from '../Agreement/assets/warning.svg'
 import { AbstainCardHeader } from '../Feed/AbstainCard'
+import LoadingRing from '@/components/LoadingRing'
```
rperez89, line 68: wondering why we removed the loadingRing component for the loading component and then we added the loadingRing component again donwn there

rperez89 (commented):
awesome work man!! just some small comments and suggestions

src/components/Garden/Feed/ProposalInfo.tsx
kamikazebr: I think that means [Constant Flow Agreement](https://docs.superfluid.finance/superfluid/protocol-overview/in-depth-overview/super-agreements/constant-flow-agreement-cfa) from Superfluid.

src/components/Garden/DisputableActionInfo.js
```diff
@@ -213,7 +213,11 @@ function usePeriod(proposal, periodEndDate) {
 }
 
 function Conviction({ proposal }) {
-  if (proposal.type === ProposalTypes.Suggestion) {
+  // TODO: handle Stream conviction data
```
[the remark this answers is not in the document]
0xGabi: Yes, we are currently showing the conviction voting data. I think that what we would like to show instead is the accrued flow rate per month.

src/components/Garden/Feed/ProposalInfo.tsx
0xGabi: Yes, @kamikazebr is right. Though I agree with @rperez89 I will change to `useSuperfluid`.

src/components/Garden/Feed/ProposalInfo.tsx
0xGabi: If I'm not mistaken because the `hideRequestedInfo` condition we only display for `Proposal` type. So I change to be more explicit.

src/components/Garden/ProposalDetail/ProposalDetail.js
0xGabi: I was trying the new Loading and add it again. Rolling back that change.

rperez89 (approved):
LGTM!

PR #893, opened 2022-11-04
Improve writting in fluid proposals

sembrestels:
It shows a more accurate message in the streaming proposals:

Screenshot from 2022-11-04 17-21-03


0xGabi (approved):
Much clear now!

PR #902, opened 2022-12-20
Add a way to force close the error modal

### LINKED ISSUE #901: Bug - can't close error modal
# Problem
The error modal can't be closed with the X button
Recording 2022-12-20 at 12 48 03

# Solution 
It could be nice if we could close this modal even if the page is in an error state in order to see what's already available

Corantin:
Closes #901 
# Before
Recording 2022-12-20 at 12 48 03
# After
Recording 2022-12-20 at 13 27 43-fix


kamikazebr (approved):
Nice catch!

PR #911, opened 2023-06-28
:adhesive_bandage: fix ipfs double slash in covenant

kamikazebr:


pglavin2 (commented):
that should do the trick

pglavin2 (approved):

PR #918, opened 2023-08-06
Fix Abstain Proposal Not opening

kamikazebr:
:bug: check proposal have link to discourse
:rotating_light: remove unecessary await


pglavin2 (approved):
Nice.

